PR #60, opened 2020-07-12
Dynamic programming and wavefront examples

pnookala:
This commit includes workflow and dataflow versions of Floyd Warshall's All Pairs Shortest Paths, Gaussian Elimination without pivoting and wavefront examples.


examples/FW-APSP/FloydIterativeKernelDF.h
```diff
@@ -0,0 +1,24 @@
...
+
+template <typename T>
+BlockMatrix<T> floyd_iterative_kernel(int block_size, BlockMatrix<T> m_ij, 
+							BlockMatrix<T> m_ik, BlockMatrix<T> m_kj) {
+	//BlockMatrix<T> bm(block_size, block_size);
+  for(int k = 0; k < block_size; ++k) {
+		//#pragma ivdep
+    for(int i = 0; i < block_size; ++i) {
+			for(int j = 0; j < block_size; ++j) {
+				m_ij(i,j,min(m_ij(i,j), m_ik(i,k) + m_kj(k,j)));
```
bosilca, line 16: This might be a valid statement, but it has no outcome right ?

examples/CMakeLists.txt
```diff
@@ -15,13 +15,34 @@ if (TARGET MADworld)
...
-	target_link_libraries(wavefront-df-mad LINK_PUBLIC ttg-mad ${MADNESS_DISABLEPIE_LINKER_FLA})
+	target_link_libraries(wavefront-df-mad LINK_PUBLIC ttg-mad ${MADNESS_DISABLEPIE_LINKER_FLAG})
+
+  # Floyd Warshall's all pairs shortest paths
+  add_executable (fw-apsp-mad madness/floyd_warshall.cc floyd_warshall.impl.h)
+  target_link_libraries(fw-apsp-mad LINK_PUBLIC ttg-mad ${MADNESS_DISABLEPIE_LINKER_FLAG})
+
+  # Floyd Warshall's all pairs shortest paths (distributed)
+  add_executable (fw-apsp-df-mad madness/floyd_warshall_df.cc floyd_warshall_df.impl.h)
+  target_include_directories(fw-apsp-df-mad PUBLIC /usr/local/opt/libomp/include/)
```
bosilca, line 36: We should not use static directories to link against (or include)

examples/CMakeLists.txt
```diff
@@ -15,13 +15,34 @@ if (TARGET MADworld)
   add_executable (t9-mad madness/t9.cc t9.impl.h)
   target_link_libraries(t9-mad LINK_PUBLIC ttg-mad ${MADNESS_DISABLEPIE_LINKER_FLAG})
 
-	#Wavefront example (shared memory)
+	#Wavefront example (workflow with shared data)
 	add_executable (wavefront-wf-mad madness/wavefront-wf.cc wavefront-wf.impl.h)
 	target_link_libraries(wavefront-wf-mad LINK_PUBLIC ttg-mad ${MADNESS_DISABLEPIE_LINKER_FLAG})
+
+  # Wavefront example (workflow with classes)
+  add_executable (wavefront-wf2-mad madness/wavefront-wf2.cc wavefront-wf2.impl.h)
```
bosilca, line 23: Are all these tests only running in the MADNESS version of the code ?

examples/FW-APSP/FloydRecursiveSerialKernelDF.h
```diff
@@ -0,0 +1,204 @@
+#ifndef FLOYD_RECURSIVE_SERIAL_KERNEL
```
bosilca, line 1: We should not maintain this file as a copy of the OMPI version. Instead we should recompile the same file twice, once with PML support and once without.

examples/CMakeLists.txt
therault: Serialization/Deserialization is not integrated in the parsec ttg driver to this day: we only use the AM system, so it only works for POD structures. This example relies on sharedpointers and serialize/deserialize, so it cannot work with parsec driver in distributed today.

examples/FW-APSP/FloydRecursiveSerialKernelDF.h
therault: I guess this comment targeted a file in Open MPI?

bosilca: That's an automatic corrector mistake. It should not have been OMPI, but OMP (like Open MP)

examples/CMakeLists.txt
pnookala: I fixed it and created a pull request. The pull request has about 12 commits related to other examples, but commits 3dd8736 and ac26f2c and can be reviewed/merged to master. Thank you.

examples/FW-APSP/FloydIterativeKernelDF.h
pnookala: This is a valid statement and it modifies the element inside the block matrix, but I changed the operator overloading to make it more intuitive. This change is part of the pull request #61

PR #62, opened 2020-07-31
Passing tuple of refs

evaleev:
@robertjharrison I think this works, but can't be certain about mrattg, please have a look


evaleev:
will be merged via #63

PR #64, opened 2020-09-14
Documentation

pnookala:
Changed README.md to include more details. Cleaned up examples.


examples/blockmatrix.h
```diff
@@ -115,28 +119,102 @@ std::ostream& operator<<(std::ostream& s, BlockMatrix<T>& m) {
...
+    m = std::shared_ptr<BlockMatrix<T>>(new BlockMatrix<T>[nb_row * nb_col], [](BlockMatrix<T>* b) { delete[] b; });
+
+    for (int i = 0; i < nb_row; i++)
+      for (int j = 0; j < nb_col; j++) {
         m.get()[i * nb_col + j] = BlockMatrix<T>(b_rows, b_cols);
+      }
   }
 
-  ~Matrix() {}
+  ~MatrixOld() {}
```
therault, line 217: Do we use MatrixOld or is it kept only for historical reasons?

examples/floyd_warshall_df.impl.h
```diff
@@ -106,7 +108,22 @@ class Initiator : public Op<int,
...
+        } else if (i == 0) {  // B function call
+          ::send<1>(Key(std::make_pair(std::make_pair(i, j), 0)), kv.second, out);
+        } else if (j == 0) {  // C function call
+          ::send<2>(Key(std::make_pair(std::make_pair(i, j), 0)), kv.second, out);
+        } else {  // D function call
+          ::send<3>(Key(std::make_pair(std::make_pair(i, j), 0)), kv.second, out);
+        }
+    });
+      
+    /*for (int i = 0; i < iterations; ++i) {
```
therault, line 126: This block is commented out: if we don't need need, let's just remove it in the examples? Or do we want to keep it for a reason?

examples/ge_df.impl.h
```diff
@@ -120,19 +121,23 @@ class Initiator
     // making x_ready for all the blocks (for function calls A, B, C, and D)
     // This triggers for the immediate execution of function A at tile [0, 0]. But
     // functions B, C, and D have other dependencies to meet before execution; They wait
-    for (int i = 0; i < iterations.value; ++i) {
-      for (int j = 0; j < iterations.value; ++j) {
+    //for (int i = 0; i < iterations.value; ++i) {
```
therault, line 124: Same here: if the two loops are commented out, why do we keep them?

examples/sw.impl.h
```diff
@@ -0,0 +1,362 @@
...
+      //std::cout << std::endl;
+   }
+   
+   // returning the data at the bottom-right as the final value
+   int final_value = X[(n-1)*n+(n-1)];
+   delete[] X;
+   return final_value;
+}
+
+/*int SW_OpenMP(const std::string &a, const std::string &b,
```
therault, line 305: Here a few commented out functions

therault (commented):
Looks good, just a few pieces of commented-out codes to check if we need to keep them or remove them

examples/sw.impl.h
pnookala: This code is the OpenMP implementation on Smith Waterman. This is kept for future use.

evaleev (approved):

PR #67, opened 2021-01-15
Parsec ttg broadcast

therault:
This PR implements a ttg_broadcast operation using the madness::archive implementation. It also fixes communication in the parsec driver when using non-trivially copyable types (also using madness:archive implementation).


examples/ttg_matrix.h
```diff
@@ -392,9 +407,10 @@ namespace ttg {
     }
 
    private:
-    data_edge_t data_edge_{};
-    shape_edge_t shape_edge_{};
-    ctl_edge_t ctl_edge_{};
+    data_edge_t data_edge_{"data_edge_"};
+    shape_edge_t read_shape_edge_{"read_shape_edge_"};
```
devreal, line 411: @evaleev Can you please take a look at the design of the two shape edges? @therault and I agreed that we should merge this branch first but I am not sure having read and write shape edges is the right way to go. Maybe there is a better way of expressing this? And then what should the constructor look like? (https://github.com/TESSEorg/ttg/pull/67/files#diff-c4b309a3be9cda758568eb949e302fe46276873919795c4e027b8ef8f001fb9eR366)

evaleev (approved):

PR #68, opened 2021-01-19
Implement ttg::World and move away from single ttg.h

devreal:
This PR contains a couple of changes I have been working on, one leading to the other:

1) Implement `ttg::World` as a slim wrapper around an implementation object, using `std::shared_ptr`. `ttg::World` should always be passed/stored by value to get the reference counting right.
2) This is the basis for lifetime tracking of op and world objects, which will now be destroyed during `ttg_finalize` to prevent destruction at the end of `main`.
3) Tear apart the monolithic `ttg.h` and put its contents into individual header files, stored in and installed into a proper directory structure. We can now pick an implementation by defining `TTG_USE_PARSEC` or `TTG_USE_MADNESS` (but not both, which is caught by the implementation) and including the main `ttg.h`. Alternatively, `ttg/parsec/ttg.h` or `ttg/madness/ttg.h` can still be included.
4) Move a bunch of examples from implementation headers to single implementation `.cc` files and select the implementation to be used in the CMake file.


therault:
We want to keep the capability to use both runtimes in the same application (at the same time).
The top header / CMake options should decide which is the default one, but the refactor should allow to explicitly include and use ::ttg::madness::op and ::ttg::parsec::op.

Including madness/ttg.h or parsec/ttg.h should not automatically import the runtime specific namespace into the ttg namespace.
There should be a separate header to do this importation.

devreal:
I made the changes we discussed at the call yesterday: using inline namespaces to import definitions from the implementation namespaces. However, things got a little ugly: even subnamespaces will clashes with namespaces in the parent namespace. Thus, we cannot have `ttg::ttg_madness::detail` (which would clash with `ttg::detail`). I renamed the implementation detail namespaces to `ttg::ttg_madness::ttg_madness_detail` and `ttg::ttg_parsec::ttg_parsec_detail`. This is far from ideal but I cannot think of anything better right now.

evaleev:
superseded by #76

PR #79, opened 2021-02-18
Fixed issues with examples, changed RandomAccess benchmark to new example format

pnookala:


examples/ge/ge_df.cc
```diff
@@ -11,16 +11,18 @@
...
 #include "GE/GEIterativeKernelDF.h"          // contains the iterative kernel
 //#include "GE/GERecursiveParallelKernel.h"  // contains the recursive but serial kernels
 #include "GE/GERecursiveSerialKernel.h"    // contains the recursive and parallel kernels
 
-#include TTG_RUNTIME_H
-IMPORT_TTG_RUNTIME_NS
+#include "ttg.h"
+using namespace ttg;
+
+#include <madness/world/world.h>
```
devreal, line 25: Why do we need the madness world header here?

examples/sw/sw.cc
```diff
@@ -218,15 +218,20 @@ auto make_sw1(const funcT& func, int block_size, const std::string &a, const std
...
+  int block_size;
+  bool verify;
+
+  if (argc < 6) {
+    problem_size = 1024;
+    block_size = 32;
+    verify = false;    
+  }
+  else {
+    problem_size = std::atoi(argv[2]);
```
devreal, line 231: Since the help output is now gone it is really hard to understand which values should be in which place of the argument list. Maybe restore the output and/or remove the parameter names (since they are not used anyway)?

devreal (changes requested):
Thank you @pnookala for looking into this. You said that GE is shared memory only so should we error out there if running on more than one rank? 

Also, we need to come up with a unified way for specifying arguments, the patchwork we have now is really confusing.

examples/ge/ge_df.cc
pnookala: This is required for using hashT from the madness namespace (for hashing pair of pairs) which is implemented in madness-src and also for using xterm_debug.

examples/sw/sw.cc
pnookala: I have kept the default values and added the help output.

PR #80, opened 2021-02-19
Remove left-over debug output during teardown

devreal:
Left-over from #76 


evaleev:
@devreal try merging master to see if this resolves the hang issues

evaleev (approved):

PR #83, opened 2021-03-12
Evaleev/feature/matrix

evaleev:


therault:
Proper design for sparse matrix where metadata computation is also part of the flow.

Requires other abstractions (tiling of metadata and others).

Merging is postponed

PR #89, opened 2021-03-24
Examples/hash functions

therault:
See commit messages:  the hash functions on tuples would create a lot of collisions for parsec's tasks hash table, generating a warning at runtime.

Most of the PR is formatting, though, because of the automatic formatting pre-commit checks...


examples/floyd-warshall/floyd_warshall.cc
```diff
@@ -195,13 +195,13 @@ class FuncA : public Op<Key,
...
-      // 	#pragma omp single
-      // 	{
-      // 		#pragma omp task
-      // 		floyd_recursive_serial_kernelA(adjacency_matrix_ttg, problem_size,
-      // 								   block_size, i_lb, j_lb, k_lb,
-      // 								   recursive_fan_out, base_size);
-      // 	}
+      //        #pragma omp single
+      //         {
+      //                 #pragma omp task
```
devreal, line 200: Do we really want 8-space indentation?

therault: Automated crap:
 - no-tab checker passes
 - clang-format finds that a line is longer than 80 characters in a // comment, introduces a carriage return, and for a reason introduces some tabs
 - iterate:
 - no-tab checker complains about tabs in files
 - Thomas does sed -i -e 's/\t/        /g' in the files, without thinking
 - all the formatting tests pass
I guess it's not only automated crap, the user is in the loop :) But clang-format shouldn't have introduced tabs.

I'll fix and try to find why clang-format sometimes introduces tabs.

devreal: Ahh yes, I realized later that it's really only in the comments where things go bad. Maybe just leave the formatting as is?

PR #91, opened 2021-04-07
Pull terminals functionality using C++ Containers

pnookala:
This PR implements pull terminals using C++ Containers as the data source for pulling data. A reference to the Container is stored inside a pull terminal and data is accessed using the "at" member function. The terminal also stores a mapping function for mapping input keys to data indexes. For remote access, a message is sent to the terminal in another process and data is received back using another send. Currently wavefront_df example uses this mechanism for pulling blocks of a matrix. The data is distributed across multiple processes in this example and it overrides the keymap by setting its own keymap based on the data distribution.

Note: Implementation for generator tasks using a variation of pure tasks will be in a separate PR.


pnookala:
Closing since this PR is old and another PR supersedes this one.

devreal:
Just for the record: containers should not be passed to an Edge directly, for various reasons (one being the inadvertent template-dependency). Instead, the same can be achieved by ref-capturing the container in a lambda and using a `std::function` that returns the element asked for...

PR #99, opened 2021-04-29
serialization refactoring

evaleev:
this _starts_ refactoring of serialization (multiple PRs will be needed to complete)

includes the following:
- `ttg-serialization` is a separate CMake library; only `ttg-parsec` depends on it
  - can support madness, boost, and cereal
- `ttg-parsec` does not use madness serialization directly any longer (but does indirectly, since mad is the only serialization supported currently by `default_data_descriptor`
- serialization is a proper unit test (Catch2 is fetch contented in)
- reveals many issues with madness serialization that will need to be addressed
- fills basic gaps in functionality of cereal (e.g. missing support for serialization of trivially-copyable types and static-sized arrays


devreal:
I glanced over it and wondered where boost and cereal are actually used. Is that something we still need to incorporate into the Madness and PaRSEC backends?

evaleev:
Not yet used anywhere. Just some initial traits etc.

CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ if (EIGEN3_FOUND)
 endif (EIGEN3_FOUND)
 # Boost
 include(FindOrFetchBoost)
+# Cereal
+include(FindOrFetchCereal)
```
devreal, line 49: Should we include `FindOrFetchCatch2` too?

ttg/ttg/serialization/boost.h
```diff
@@ -0,0 +1,35 @@
+//
+// Created by Eduard Valeyev on 5/3/21.
```
devreal, line 2: Should we use proper copyright headers? (might be a question beyond the files added here)

devreal (approved):
I guess we should overload `default_data_descriptor` for cereal, boost, and catch2 (in whatever priority). I'm fine with merging the current state though. I will rebase my iovec implementation and we take it from there.

CMakeLists.txt
evaleev: We can discuss how to organize unit tests .. right now the only dir with unit tests loads catch2

evaleev:
> I guess we should overload `default_data_descriptor` for cereal, boost, and catch2 (in whatever priority). I'm fine with merging the current state though. I will rebase my iovec implementation and we take it from there.

hold on, I will do more PRs. `default_data_descriptor` will hide the serialization details (behaving ttg archive types), so single instance will be sufficient. Let's discuss how to handle iovec on Friday.

ttg/ttg/serialization/boost.h
evaleev: yes, should discuss on Friday

PR #101, opened 2021-05-12
multiple serialization backends

evaleev:
- Adds support for Boost-serializable types
- custom TTG boost archives for efficient (i.e. no class info, pointer tracking, etc.) serialization
- IOVEC archives .. Boost-only, but logic easy to replicate in MADNESS archives
- Parsec backend by default now uses boost serialization if it's available, otherwise reverts to MADNESS (if available)


examples/spmm/spmm.cc
```diff
@@ -23,129 +24,27 @@ using namespace ttg;
 
 #if defined(BLOCK_SPARSE_GEMM) && defined(BTAS_IS_USABLE)
 using blk_t = btas::Tensor<double>;
+
+// declare btas::Tensor serializable by Boost
+#include "ttg/serialization/backends/boost.h"
+namespace ttg::detail {
+  template <typename Archive>
+  inline static constexpr bool is_boost_serializable_v<Archive, blk_t> = is_boost_archive_v<Archive>;
```
devreal, line 32: Why not simplify this to 

```
inline static constexpr bool is_boost_serializable_v<blk_t> = true;
```

Why is that dependent on the Archive?

examples/wavefront/wavefront-wf2.cc
```diff
@@ -75,13 +77,20 @@ class BlockMatrix {
   }
 
   T& operator()(int row, int col) { return m_block.get()[row * _cols + col]; }
+  const T& operator()(int row, int col) const { return m_block.get()[row * _cols + col]; }
 
   template <typename Archive>
-  void serialize(Archive& ar) {}
+  void serialize(Archive& ar) {
+    std::abort();
```
devreal, line 84: Use `ttg::abort` instead

examples/wavefront/wavefront-wf2.cc
```diff
@@ -75,13 +77,20 @@ class BlockMatrix {
...
+  }
+  template <typename Archive>
+  void serialize(Archive& ar, const unsigned int) {
+    std::abort();
```
devreal, line 88: Use `ttg::abort` instead

ttg/ttg/serialization/backends/boost/archive.h
```diff
@@ -0,0 +1,295 @@
...
+}  // namespace ttg::detail
+
+// for some reason need to use array optimization for the base as well ... dispatch to optimized version in
+// array_wrapper.hpp:serializer(ar,version) for some reason uses Archive::base_type using apple clang 12.0.5.12050022
+#define BOOST_SERIALIZATION_USE_ARRAY_OPTIMIZATION_FOR_THIS_AND_BASE(x) \
+  BOOST_SERIALIZATION_USE_ARRAY_OPTIMIZATION(x);                        \
+  BOOST_SERIALIZATION_USE_ARRAY_OPTIMIZATION(x::base_type);
+
+BOOST_SERIALIZATION_REGISTER_ARCHIVE(ttg::detail::boost_counting_oarchive);
+BOOST_SERIALIZATION_USE_ARRAY_OPTIMIZATION_FOR_THIS_AND_BASE(ttg::detail::boost_counting_oarchive)
```
devreal, line 283: Missing `;` at the end of the line?

ttg/ttg/serialization/backends/cereal.h
```diff
@@ -0,0 +1,98 @@
...
+  inline constexpr bool is_cereal_user_serializable_v = is_cereal_user_serializable<Archive, T>::value;
+
+  template <typename T, class = void>
+  struct is_cereal_user_buffer_serializable : std::false_type {};
+
+#ifdef TTG_SERIALIZATION_SUPPORTS_CEREAL
+  template <typename T>
+  struct is_cereal_user_buffer_serializable<
+      T, std::enable_if_t<is_cereal_user_serializable_v<cereal::BinaryInputArchive, T> ||
+                          is_cereal_user_serializable_v<cereal::BinaryOutputArchive, T>>> : std::true_type {};
```
devreal, line 89: Shouldn't we test with `&&` here? We need support for both input and output archive from T, no?

ttg/ttg/serialization/data_descriptor.h
```diff
@@ -125,6 +120,92 @@ namespace ttg {
...
+      cereal::BinaryOutputArchive oa(os);
+      oa << (*(T *)object);
+      return sbuf.size();
+    }
+
+    /// object --- obj to be serialized
+    /// chunk_size --- inputs max amount of data to output, and on output returns amount actually output
+    /// pos --- position in the input buffer to resume serialization
+    /// buf[pos] --- place for output
+    static uint64_t pack_payload(const void *object, uint64_t chunk_size, uint64_t pos, void *_buf) { abort(); }
```
devreal, line 196: Why abort here?

ttg/ttg/serialization/std/list.h
```diff
@@ -0,0 +1,43 @@
...
+// MADNESS supports std::list serialization by default
+#endif
+
+#ifdef TTG_SERIALIZATION_SUPPORTS_BOOST
+#include <boost/serialization/list.hpp>
+
+namespace ttg::detail {
+  template <typename Archive, typename T, typename A>
+  inline static constexpr bool is_stlcontainer_boost_serializable_v<Archive, std::list<T, A>> =
+      is_boost_serializable_v<Archive, T>&& is_boost_serializable_v<Archive, A>;
```
devreal, line 21: Nitpick: the formatting here is weird:

```suggestion
      is_boost_serializable_v<Archive, T> && is_boost_serializable_v<Archive, A>;
```

examples/floyd-warshall/floyd_warshall.cc
```diff
@@ -18,14 +18,14 @@
...
 // #include "FloydRecursiveParallelKernel.h" // contains the recursive and parallel kernels
 
-// needed for madness::hashT and xterm_debug
-#include <madness/world/world.h>
-
 #include "ttg.h"
 using namespace ttg;
 
+#include "ttg/serialization.h"
+#include "ttg/serialization/std/pair.h"
```
devreal, line 25: Why not include `std/pair.h` et al. in `serialization.h? I need to remember to include it, otherwise compilation fails, right?

ttg/ttg/util/version.cc
```diff
@@ -1,7 +1,5 @@
-#include "../../ttg/util/version.h"
+#include "ttg/util/version.h"
 
 namespace ttg {
-  std::array<int,3> version() {
-    return {0,1,0};
-  }
-}
\ No newline at end of file
+  std::array<int, 3> version() { return {0, 1, 0}; }
```
devreal, line 4: For the record: I'm not a fan of single-line-functions-everywhere. This change does not improve readability :/ (there are similar changes in this PR for which this is true too)

ttg/ttg/serialization/data_descriptor.h
```diff
@@ -125,6 +120,92 @@ namespace ttg {
...
+      T, std::enable_if_t<(!std::is_trivially_copyable<T>::value && !detail::is_madness_buffer_serializable_v<T> &&
+                           detail::is_boost_buffer_serializable_v<T>) ||
+                          (!detail::is_madness_user_buffer_serializable_v<T> &&
+                           detail::is_boost_user_buffer_serializable_v<T>)>> {
+    static constexpr const bool serialize_size_is_const = false;
+
+    static uint64_t payload_size(const void *object) {
+      ttg::detail::boost_counting_oarchive oa;
+      oa << (*(T *)object);
+      return oa.streambuf().size();
```
devreal, line 142: We need to serialize the object to find out it's serialized size? That means we serialize every object twice?

CMakeLists.txt
```diff
@@ -32,6 +32,11 @@ include(ExternalDependenciesVersions)
 include(LoadFetchContent)
 include(AddCustomTargetSubproject)
 
+########################################
+#### user-defined configuration options
+########################################
+option(TTG_PARSEC_USE_BOOST_SERIALIZATION "Whether to select Boost serialization methods in PaRSEC backend" ON)
```
devreal, line 38: What is the use of this option? Is there a similar option for MADNESS? Isn't support for madness/boost/cereal transparent to the backend?

devreal (changes requested):
Looks good, some comments though

CMakeLists.txt
therault: That was the consensus... But the first serialization does not write in memory, so costs way less, and if we're lucky the serialized object fits in some levels of cache, so the sequence serialize_to_get_the_size() ; allocate() ; serialize() may not cost significantly more than serialize() by itself. Alternative is to work with re-dimensionable / re-allocatable objects which is not too good either...

ttg/ttg/serialization/data_descriptor.h
devreal: Ahh yes, I didn't realize that the madness descriptor does the same. But why does the first serialize not perform stores? The object could be on the stack and not valid after the call so the archive has to make a copy.

We could allow the user to give us a hint at the size, e.g., by overloading `ttg::serialization::user_payload_size(T&)`. If that's not available we fall back to the double serialization approach.

therault: Yes, agreed, this is only the default implementation, when we can't do better.

Note that in `payload_size`, we use a `ttg::detail::boost_counting_oarchive`, and not a `ttg::detail::boost_oarchive`. `boost_counting_oarchive` should have an empty storage, and only do the counting.

evaleev: @devreal no way in general to compute data structure a priori in C++ except for trivial cases. Thus unfortunately must touch all data twice unless we want to make the 2-step mechanism only necessary for objects that exceed fixed size ... i.e. we could serialize into fixed-size buffer (user-configurable size, presumably as argument to ttg_initialize or smith) and if a given object exceeds its size we could use exceptions to signal that and invoke the 2-step mechanism.

I don't know enough about runtime performance impact of having exceptions in this path. But should be manageable. This will make serialization optimally fast for all but very large objects.

ttg/ttg/util/version.cc
evaleev: looks like auto-clang-formatting by the git-hook ... will have a look.

examples/floyd-warshall/floyd_warshall.cc
evaleev: include-what-you-use. although that train left the station eons ago (we use variadics and heavy metaprogramming everywhere in TTG) it seems to be unreasonable to include support for serialization of all `std` types

ttg/ttg/serialization/std/list.h
evaleev: indeed ... let me check why clang-format does that (or if it somehow was not invoked by git-hooks)

ttg/ttg/serialization/backends/cereal.h
evaleev: If user provided serialization method for either we should insist on using user serializability. If only one was provided this will likely lead to compile-time error, or the user indeed wanted this to happen

examples/spmm/spmm.cc
evaleev: by definition serializability is Archive-dependent. e.g. user could limit serialization to specific Archive (or class of Archives), could limit it to output archive only, etc.

ttg/ttg/serialization/std/list.h
evaleev: missing space is indeed what the clang-format 11.1.0 on my Mac produces with our `.clang-format` ...

evaleev:
@devreal the clang-format issues you pointed out led to a black hole. let's postpone these issues to another PR

PR #103, opened 2021-05-20
Extend the examples/spmm example to work with larger matrices

therault:
Add two options to examples/spmm:
  - -mm /path/to/a/MatrixMarket/sparse/matrix
  - -rmat N[:E[:a[:b[:c[:d]]]]] use a random graph generator
    in Boost to build the connectivity matrix of that graph

If either -mm or -rmat are used, we compute C = A * A, if A is
square, or C = A * A^T, if A is not square, where A
is the matrix provided / generated. Some rudimentary
statistics about the matrices (size and number of non zero),
and the time to compute the product.

If neither -mm or -rmat are specified, the example continues to
use the previous behavior.


evaleev:
@therault merge master in and CI should then pass

devreal:
Automatic formatting turned this PR into a train wreck. The salvage is now at #136. Closing.

PR #104, opened 2021-05-23
Evaleev/fix/clang format style

evaleev:


therault:
Decided to close, as we cannot agree on automatic formatting yet.

PR #120, opened 2021-06-24
shallow copy btas tiles

evaleev:
bspmm uses shallow-copy BTAS tensor


devreal:
There is an awful lot of formatting changes in the parsec ttg header, which are unrelated to the changes in this branch :/

PR #125, opened 2021-07-02
fix non-POSIX-standard function def in shell scripts

milthorpe:
The docker-build.sh script failed for me on Ubuntu 20.04 with:
`$ bin/docker-build.sh amd64
bin/docker-build.sh: 78: function: not found`
This is due to the use of non-POSIX-standard bash syntax for function definition, which this patch fixes.


evaleev (approved):

evaleev:
Thanks, @milthorpe !

PR #132, opened 2021-08-10
PaRSEC backend: rename my_op_t and store key inline

devreal:
`my_op_t` is replaced by `parsec_ttg_op_t` and templated on the key type. `parsec_ttg_op_t` is derived from `parsec_ttg_op_base_t` which can be used in context where the key type is unknown or not needed. Storing the key inline avoids memory in the critical path taken when handling an input to a task that was not discovered yet.


devreal:
This PR has evolved quite a bit and now contains enhancements to the PaRSEC backend aimed at reducing the overhead of creating tasks once discovered, esp. in dealing with each Op's central hash table used to track discovered but not yet eligible tasks.

devreal:
Superseded by https://github.com/TESSEorg/ttg/pull/140, closing.

PR #137, opened 2021-08-26
Performance Improvements for SPMM

devreal:
This PR adds small improvements to the SPMM implementation, including:

- Marking MultiplyAdd's input tiles as const eliminates copies in the PaRSEC backend
- Split allocation and copying of tiles in the result matrix
- Add task priorities using the number of remaining operations on the tile

### DIFF
--- a/examples/spmm/spmm.cc
+++ b/examples/spmm/spmm.cc
@@ -301,9 +301,9 @@ class SpMM {
   };  // class BcastA
 
   /// multiply task has 3 input flows: a_ijk, b_ijk, and c_ijk, c_ijk contains the running total
-  class MultiplyAdd : public Op<Key<3>, std::tuple<Out<Key<2>, Blk>, Out<Key<3>, Blk>>, MultiplyAdd, Blk, Blk, Blk> {
+  class MultiplyAdd : public Op<Key<3>, std::tuple<Out<Key<2>, Blk>, Out<Key<3>, Blk>>, MultiplyAdd, const Blk, const Blk, Blk> {
    public:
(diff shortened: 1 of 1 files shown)


examples/spmm/spmm.cc
```diff
@@ -195,14 +195,16 @@ class Write_SpMatrix : public Op<Key<2>, std::tuple<>, Write_SpMatrix<Blk>, Blk>
...
     if (ttg::tracing()) {
       auto &w = get_default_world();
       ttg::print(w.rank(), "/", reinterpret_cast<std::uintptr_t>(pthread_self()), "spmm.impl.h Write_SpMatrix wrote {",
                  key[0], ",", key[1], "} = ", baseT::template get<0>(elem), " in ", static_cast<void *>(&matrix_),
                  " with mutex @", static_cast<void *>(&mtx_), " for object @", static_cast<void *>(this));
     }
-    matrix_.insert(key[0], key[1]) = baseT::template get<0>(elem);
+    auto entry = matrix_.insert(key[0], key[1]);
+    mtx_.lock();
+    entry = baseT::template get<0>(elem);
```
evaleev, line 207: For posterity: `entry` can be invalidated by inserts happening since lock release (if such insert triggers internal buffer resizing)

devreal: Mhh, that's unfortunate. That makes the critical path extremely much longer... Any idea how to work around this?

evaleev: should not be critical for shallow copy tiles, #120

PR #140, opened 2021-09-02
Improvements to the PaRSEC backend

devreal:
- Template the task structure on the key type and the potential number of streaming terminals
- Use the local thread's execution stream, if available. Previously, all threads would hammer the same execution stream.
- Avoid locking hash tables where possible to reduce some congestion
- In a broadcast, submit tasks to `__parsec_schedule` at once

### DIFF
--- a/ttg/ttg/parsec/ttg.h
+++ b/ttg/ttg/parsec/ttg.h
@@ -68,6 +68,7 @@ int parsec_add_fetch_runtime_task(parsec_taskpool_t *tp, int tasks);
 
 namespace ttg_parsec {
   static __thread parsec_task_t *parsec_ttg_caller;
+  static __thread parsec_execution_stream_t *parsec_ttg_es;
 
   typedef void (*static_set_arg_fct_type)(void *, size_t, ttg::OpBase *);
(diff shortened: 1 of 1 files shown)


ttg/ttg/parsec/ttg.h
```diff
@@ -68,6 +68,7 @@ int parsec_add_fetch_runtime_task(parsec_taskpool_t *tp, int tasks);
 
 namespace ttg_parsec {
   static __thread parsec_task_t *parsec_ttg_caller;
+  static __thread parsec_execution_stream_t *parsec_ttg_es;
```
evaleev, line 71: - `__thread` is nonstandard, switch to `thread_local`: http://en.cppreference.com/w/cpp/language/storage_duration
- probably do not need `static` to indicate internal linkage here? `static` will make different variables in different object files that include `parsec/ttg.h`. Presumably these should just be per thread, not per thread + per object file. So I suggest `inline thread_local` (see https://en.cppreference.com/w/cpp/language/inline)

evaleev (changes requested):

evaleev (approved):

PR #146, opened 2021-09-07
SPMM: stay within index matrix bounds

devreal:
`std::vector::operator[]` does not perform bounds checks so occasionally went oob without noticing, unless we ended up in some strange segfaults.

@evaleev  I actually don't know if this is the proper fix. It might as well just fix the symptoms, not the cause (an actual error in the index calculations...)


evaleev (approved):

evaleev:
thanks!

PR #154, opened 2021-09-09
SPMM: scramble the hash using a randomly chosen prime number

devreal:
The default keymap uses the hash for process selection.
The indeces in Eigen are the index of the tile's first element
in the matrix so the key directly depends on the tile size.
If the tile size is a multiple of the number of processes all
tasks end up on one process... We better scramble the hash!


evaleev:
@therault this is probably obsolete thanks to #157 ?

devreal:
I don't think this is needed anymore, closing.

PR #157, opened 2021-09-10
bspmm-distributed

therault:
- Generate the matrices distributed
- Enable a -c option to set the number of cores


examples/spmm/spmm.cc
```diff
@@ -946,6 +1035,39 @@ int main(int argc, char **argv) {
...
+      else {
+        if (0 == mpi_rank) {
+          std::cerr << P << "x" << Q << " is not a valid process grid -- bailing out" << std::endl;
+          MPI_Abort(MPI_COMM_WORLD, EXIT_FAILURE);
+        }
+      }
+    }
+
+    std::string seedStr(getCmdOption(argv, argv + argc, "-s"));
+    unsigned int seed = parseOption(seedStr, 0);
```
devreal, line 1057: There is a problem here: `parseOption` uses `std::stoi` to parse the string but the provided seed is `unsigned int`, potentially leading to an out-of-range exception if passed back in. One solution is to use `std::stol` in `parseOption` :)

evaleev (approved):

PR #158, opened 2021-09-16
Spmm reduce gemms

therault:
Dot Not Merge: this branch is experimental, to see if we can look at the performance of a reducer-based GEMM.


therault:
Used as a variant when exploring different algorithms for a paper. Not merged.

PR #159, opened 2021-09-16
Add region probe infrastructure

devreal:
The region probes are divided into three levels:
- User probes: defined by the user using ttg::user_probe; compile with -DTTG_ENABLE_USER_PROBES=1 to enable.
- Task probes: task execution instrumentation using names provided to ttg::Op; compile with -DTTG_ENABLE_TASK_PROBES=1 to enable.
- Internal probes: all the gory runtime-internal details; compile with -DTTG_ENABLE_INTERNAL_PROBES=1 to enable.

At the moment, the latter two are only available in the PaRSEC backend.
User probes can be defined any time and work with any backend that uses PaRSEC.

To enable tracing in PaRSEC, users have to export the following two
environment variables:
- PARSEC_MCA_mca_pins=task_profiler
- PARSEC_MCA_profile_filename=<filename> (files must not exist)

Note that in the wake of the instrumentation, I also encapsulated some code snippets to avoid (and reducing existing) code duplication. 


ttg/ttg/util/region_probe.h
```diff
@@ -0,0 +1,264 @@
+#ifndef TTG_REGION_PROBE_H
+#define TTG_REGION_PROBE_H
+
+#include <string>
+#include <vector>
+#include <cassert>
+
+#include <parsec/profiling.h>
```
evaleev, line 8: should this be protected by ifdefs to avoid having to introduce dependence of `ttg` target on parsec targets. Alternatively we can make parsec a hard dependence of TTG, but that may be undesirable long term.

devreal: I don't think we have a `HAVE_PARSEC` preprocessor flag right now and both the madness and parsec backends provide access to the parsec header/libs. If we create a backend that does not depend on PaRSEC we'll have to revisit that.

PR #169, opened 2021-10-14
PaRSEC: with reducers, see if the current task already has a copy of the data

devreal:


therault:
Functionality has been fixed in previous PR

PR #186, opened 2021-10-29
Devirtualize WorldImpl::size and rank

devreal:
Also avoid MPI calls in the PaRSEC backend (I was sure I had fixed that somewhere...)


therault (approved):

evaleev (changes requested):

devreal:
@evaleev Done, please check again. This is only for base classes so it shouldn't concern users.

evaleev (approved):

PR #188, opened 2021-11-01
MRA example using streaming terminals

pnookala:
Added MRA example using streaming terminals, removed the dependency on MKL and incorporated terminology changes from master branch.


pnookala:
Commit history not clean due to changes for the paper, will create separate branch for the example.

PR #190, opened 2021-11-05
[WIP] UxT-Reducer streaming terminals

devreal:
Implement UxT reduction streaming terminals. The types of edge inputs and the types of the `op` arguments (if different) are passed as tuples instead of variadic template arguments. This allows us to convert between U and T and to use different U and T for the reducer signature. 

This is WIP and so far only contains a draft of the interface in the PaRSEC backend, no UxT reducer tests yet. 

### DIFF
--- a/examples/floyd-warshall/floyd_warshall.cc
+++ b/examples/floyd-warshall/floyd_warshall.cc
@@ -95,8 +95,7 @@ std::ostream& operator<<(std::ostream& s, const Control& ctl) {
 
 class Initiator : public TT<int, std::tuple<Out<Key, Control>, Out<Key, Control>, Out<Key, Control>, Out<Key, Control>>,
                             Initiator> {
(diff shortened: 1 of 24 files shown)


devreal:
@evaleev I adapted the madness backend to the new tuple-based TT type list. Please take a look at it if you get a chance.

@pnookala Could you take a look at how the MRA benchmark could work with the new interface? It would be a good test case to try out before merging :) In short, you can use different types in the input edge and the op argument and provide a reducer that "converts" between the two.

devreal:
Pushed some fixes. I still cannot compile spmm with the madness backend and I have a hard time figuring out where the template void-voodoo goes haywire...

therault:
This approach is problematic, because it removes the compile-time checking capabilities of reducers.

It is replaced with PR #229

PR #191, opened 2021-11-05
Pull terminals functionality using containers.

pnookala:
This PR implements pull terminals using C++ Containers as the data source for pulling data. A reference to the Container is stored inside a pull terminal and data is accessed using the "at" member function. The terminal also stores a mapping function for mapping input keys to data indexes. For remote access, a message is sent to the terminal in another process and data is received back using another send. Currently wavefront_df example uses this mechanism for pulling blocks of a matrix. The data is distributed across multiple processes in this example and it overrides the keymap by setting its own keymap based on the data distribution.

Note: This PR doesn't include the functionality for updating the data in the container, it will in another PR.


bosilca:
This breaks the dataflow, because it allows any task to fetch data directly from a container instead from a predecessor. As there is no mechanism to ensure data consistency in these containers, the opportunities for incorrect code increases, and the responsibility for ensuring consistency is now on the user.

This is not what I understood from our original discussion about pull terminals. In my understanding these edges were still connected to tasks, but these tasks creation and execution was delayed until they were needed by one of their successors. This does indeed means that for pull terminal we need to know the predecessors of a task, to be able to create it and get it's data. Thus, it will look as if the DAG could be discovered in both directions, the normal flow of data from a task to it's predecessors being normal creation of the DAG, but some parts need to be remotely triggered in order to generate this flow.

We could implement the approach I describe above with normal push terminals, but it will require to invert a portion of the DAG (what was a predecessor would now become a successor) via an additional TT interposed between the them.

pnookala:
This PR contains a specialization of pull terminals concept where data from a given container can be accessed without creating a task. This is particularly useful for scenarios where a task requires additional data for the computation, and which is not really a data dependency on the predecessor task. For example, in Floyd Warshall's implementation, without pull terminals, we would need a task that would instantiate all the N^2 tasks with the block of data as input. Using the Container approach, this block can be obtained from the data structure when the task gets created during the execution of the DAG or other data dependencies on the task have been satisfied.

The pull terminals using predecessor tasks for reverse processing of the DAG will be part of another PR.

bosilca:
I understand, but I would be very careful with the usage of this capability because we have no way to guarantee data consistency.

ttg/ttg/base/tt.h
```diff
@@ -166,6 +177,13 @@ namespace ttg {
         }
       }
     }
+    
+    bool set_lazy_pull_instance(bool value) {
```
devreal, line 181: Can you please document what a lazy pull terminal is?

ttg/ttg/edge.h
```diff
@@ -59,12 +73,17 @@ namespace ttg {
 
       void try_to_connect_new_out(TerminalBase *out) const {
         assert(out->get_type() != TerminalBase::Type::Write);  // out must be an In<>
-        for (auto in : ins)
-          if (in && out) in->connect(out);
+        if (out->is_pull_terminal) {
+          out->connect_pull_nopred(out);
```
devreal, line 77: Can an Out Terminal be a pull terminal? I'm confused

ttg/ttg/edge.h
```diff
@@ -83,7 +102,15 @@ namespace ttg {
     static_assert(std::is_same_v<valueT, std::decay_t<valueT>>,
                   "Edge<keyT,valueT> assumes valueT is a non-decayable type");
 
-    Edge(const std::string name = "anonymous edge") : p(1) { p[0] = std::make_shared<EdgeImpl>(name); }
+    //Edge(const std::string name = "anonymous edge") : p(1) { p[0] = std::make_shared<EdgeImpl>(name); }
+    Edge(const std::string name = "anonymous edge", bool is_pull = false) : p(1) {
```
devreal, line 106: What happens if I create an edge and set `is_pull = true`?

ttg/ttg/edge.h
```diff
@@ -103,6 +130,10 @@ namespace ttg {
       return result;
     }
 
+    bool is_pull_edge() const {
+      return p.at(0)->is_pull_edge;
```
devreal, line 134: Should/can we handle the case where an Edge represents multiple edges? (https://github.com/TESSEorg/ttg/pull/191/files#diff-428ea16d89d546c44e669e55118f6af786a3339aa972f5aafcd17bc03d40bb48R116)

ttg/ttg/terminal.h
```diff
@@ -11,8 +11,80 @@
...
+        ContainerWrapper(T &t, mapperT &&mapper,
+                         keymapT &&keymap) : get([&t,&mapper](keyT const &key) {
+                                                  if constexpr (!std::is_class_v<T> && std::is_function_v<T>) {
+                                                      auto k = mapper(key);
+                                                      return t(k); //Call the user-defined lambda function.
+                                                    }
+                                                  else
+                                                    {
+                                                      auto k = mapper(key);
+                                                      using key_type = typename T::key_type;
```
devreal, line 43: `key_type` is not used here

ttg/ttg/terminal.h
```diff
@@ -11,8 +11,80 @@
...
+      std::function<size_t (keyT const& key)> owner = nullptr;
+
+      ContainerWrapper() = default;
+
+      template<typename T, typename mapperT, typename keymapT, std::enable_if_t<!std::is_same<std::decay_t<T>,
+                                                          ContainerWrapper>{}, bool> = true>
+        //Store a pointer to the user's container in std::any, no copies
+        ContainerWrapper(T &t, mapperT &&mapper,
+                         keymapT &&keymap) : get([&t,&mapper](keyT const &key) {
+                                                  if constexpr (!std::is_class_v<T> && std::is_function_v<T>) {
```
devreal, line 36: Do you mean `std::is_invocable_v` instead of `std::is_function_v`? https://en.cppreference.com/w/cpp/types/is_function

devreal: I think we should capture `t` and `mapper` by value. Otherwise bad things happen if they go out of scope (e.g., at the end of a `make_my_task` function)

ttg/ttg/terminal.h
```diff
@@ -11,8 +11,80 @@
...
+                                                    return keymap(idx);
+                                                  })
+        {
+         //Validate mapper and keymap function arguments
+         using mapper_args_t = boost::callable_traits::args_t<mapperT>;
+         constexpr auto num_args = std::tuple_size<mapper_args_t>::value;
+         static_assert(num_args == 1);
+         static_assert(std::is_same_v<typename std::tuple_element<0, mapper_args_t>::type, const keyT &>,
+                       "argument of Mapper must be const keyT& (unless keyT = void)");
+         using km_args_t = boost::callable_traits::args_t<keymapT>;
```
devreal, line 59: That won't work with generic lambdas. Maybe `std::is_invocable_r` will work as well?

devreal: Actually, we probably don't need to validate the functions here, the compiler will balk if the types don't pan out...

devreal (changes requested):
Couple questions I had while looking through the code.

ttg/ttg/edge.h
pnookala: We should probably limit pull terminals to be composed using single edges only.

ttg/ttg/madness/ttg.h
```diff
@@ -1025,6 +1096,7 @@ namespace ttg_madness {
 
       connect_my_inputs_to_incoming_edge_outputs(std::make_index_sequence<numins>{}, inedges);
       connect_my_outputs_to_outgoing_edge_inputs(std::make_index_sequence<numouts>{}, outedges);
+      register_input_callbacks(std::make_index_sequence<numins>{});
```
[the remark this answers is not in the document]
pnookala: Oversight while merging the changes. Will remove it.

ttg/ttg/edge.h
devreal: I remember we talked about that a while back. It would be nice to be able to merge pull and regular terminals but it was not immediately clear how to make sure when to pull from a terminal and when to wait for a value. A predicate might solve that issue (akin to making `is_pull_edge` dependent on a `key`). We should at least throw an exception in the multi-Edge constructor if one of the edges is a pull terminal.

ttg/ttg/madness/ttg.h
```diff
@@ -323,6 +325,52 @@ namespace ttg_madness {
...
+            args->counter--;
+          }
+        }
+      }
+    }
+
+    template <std::size_t i, typename Key>
+    void get_terminal_data(const int owner,
+                           const Key &key) {
+      if (owner != world.rank()) {
```
devreal, line 355: Is there any case where `get_terminal_data` is called with `owner == world.rank()`? I only see the case here https://github.com/TESSEorg/ttg/pull/191/files#diff-90cb53d7fdb7d2978ed2afc294cb0ba485aaa2b246183c8840bedfbf5d14a0e5R333

pnookala: Yes, in case of local data, owner will be same as world.rank(). Wavefront example has data distributed across processes, so some data is local and some is remote.

ttg/ttg/edge.h
pnookala: Agree. I added a check in the multi-edge constructor.

ttg/ttg/util/meta.h
```diff
@@ -427,6 +426,21 @@ namespace ttg {
       template <typename... valueTs>
       using input_reducers_t = typename input_reducers<valueTs...>::type;
 
+      //Callback type for pull Ops.
+      template<typename Key, typename Enabler = void>
+      struct invoke_callback;
```
devreal, line 431: Mhh, these don't seem to be used anywhere. Still needed?

ttg/ttg/terminal.h
```diff
@@ -11,8 +11,66 @@
...
+      std::function<valueT (keyT const& key)> get = nullptr;
+      std::function<size_t (keyT const& key)> owner = nullptr;
+
+      ContainerWrapper() = default;
+
+      template<typename T, typename mapperT, typename keymapT, std::enable_if_t<!std::is_same<std::decay_t<T>,
+                                                          ContainerWrapper>{}, bool> = true>
+        //Store a pointer to the user's container in std::any, no copies
+        ContainerWrapper(T &t, mapperT &&mapper,
+                         keymapT &&keymap) : get([&t,&mapper](keyT const &key) {
```
devreal, line 35: We should not capture `mapper` by reference. If we do, this code snippet will end up with a dangling reference:
```
ContainerWrapper(vector, [](const Key& key){ return key.x; }, ...);
```

The lambda may be destroyed after the call, it's a temporary after all.

One solution would be:
```suggestion
                         keymapT &&keymap) : get([&t,mapper = std::forward<mapperT>(mapper)](keyT const &key) {
```

Here mapper is caught by value (which is what we want, and it's typically not expensive). 

The same is true for the `owner` below.

ttg/ttg/terminal.h
```diff
@@ -11,8 +11,66 @@
...
+                                                }),
+                                            owner([&t, &mapper, &keymap](keyT const &key) {
+                                                    auto idx = mapper(key); //Mapper to map task ID to index of the data structure.
+                                                    return keymap(idx);
+                                                  })
+        {}
+    };
+
+    template <typename valueT> struct ContainerWrapper<void, valueT> {
+      std::function<std::nullptr_t ()> get = nullptr;
```
devreal, line 55: These overloads are missing proper construction.

examples/CMakeLists.txt
```diff
@@ -27,8 +27,9 @@ endif (TARGET MADworld)
 add_ttg_executable(wavefront-wf wavefront/wavefront-wf.cc SINGLERANKONLY)
 add_ttg_executable(wavefront-wf2 wavefront/wavefront-wf2.cc SINGLERANKONLY)
 add_ttg_executable(wavefront-df wavefront/wavefront-df.cc)
+add_ttg_executable(wavefront-pull wavefront/wavefront-pull.cc LINK_LIBRARIES MADworld)
 add_ttg_executable(fw-apsp floyd-warshall/floyd_warshall.cc LINK_LIBRARIES MADworld SINGLERANKONLY)
-
+add_ttg_executable(simplegenerator simplegenerator/simplegenerator.cc RUNTIMES "mad")
```
devreal, line 32: Is the `simplegenerator` still madness specific?

ttg/ttg/terminal.h
```diff
@@ -230,6 +292,19 @@ namespace ttg {
       return get_connections();
     }
 
+    template <typename Key = keyT, typename Value = valueT>
+    std::enable_if_t<meta::is_none_void_v<Key,Value> && std::is_same_v<Value,std::remove_reference_t<Value>>,void>
+    send_to(const Key &key, Value &&value, std::size_t i)
```
devreal, line 297: Is this used anywhere?

pnookala: All the unused code is leftover from pure tasks implementation for pull terminals. I will remove it.

therault:
@pnookala will resolve conflicts and ping the reviewers for merge.

ttg/ttg/parsec/ttg.h
```diff
@@ -1429,6 +1511,11 @@ namespace ttg_parsec {
           parsec_list_item_ring_push_sorted(&(*task_ring)->super, &task->parsec_task.super,
                                             offsetof(parsec_task_t, priority));
         }
+      } else if constexpr (!ttg::meta::is_void_v<keyT>) {
```
devreal, line 1514: The PaRSEC backend doesn't support void keys. Is that general policy or do we need to fix that? Do we make sure in the upper layers that pull terminals are not used with void keys?

ttg/ttg/util/meta.h
```diff
@@ -3,7 +3,6 @@
 
 #include <functional>
 #include <type_traits>
-
```
devreal, line 6: I'd prefer to have that blank line :D

ttg/ttg/terminal.h
```diff
@@ -11,8 +11,66 @@
...
+      std::function<valueT (keyT const& key)> get = nullptr;
+      std::function<size_t (keyT const& key)> owner = nullptr;
+
+      ContainerWrapper() = default;
+
+      template<typename T, typename mapperT, typename keymapT, std::enable_if_t<!std::is_same<std::decay_t<T>,
+                                                          ContainerWrapper>{}, bool> = true>
+        //Store a pointer to the user's container in std::any, no copies
+        ContainerWrapper(T &t, mapperT &&mapper,
+                         keymapT &&keymap) : get([&t,mapper = std::forward<mapperT>(mapper)](keyT const &key) {
```
devreal, line 35: I'm a little worried that if we don't forward `t` we risk losing the lambda. And why do we only forward the mapper and not the keymap? Maybe it should be:

```C
ContainerWrapper(T &&t, mapperT &&mapper,
                         keymapT &&keymap) : get([t = std::forward<T>(t),
                                                  mapper = std::forward<mapperT>(mapper)
                                                  keymap = std::forward<keymapT>(keymap)](keyT const &key) {
```

devreal (changes requested):
A few minor nits, otherwise I think it's ready to go. 

But: aren't pull terminals meant to make examples like the wavefront simpler? It still seems terribly complex with many TTs and edges...

ttg/ttg/parsec/ttg.h
pnookala: Added a static_assert in Edge constructor to take care of this.

ttg/ttg/terminal.h
```diff
@@ -11,8 +11,66 @@
...
+     * Elements of the data structure can be accessed using get method, which calls the at method of the Container.
+     * keyT - taskID
+     * valueT - Value type of the Container
+    */
+    template<typename keyT, typename valueT>
+    struct ContainerWrapper {
+      std::function<valueT (keyT const& key)> get = nullptr;
+      std::function<size_t (keyT const& key)> owner = nullptr;
+
+      ContainerWrapper() = default;
```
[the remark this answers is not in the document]
pnookala: Done.

devreal (approved):

PR #193, opened 2021-11-08
introduced ttg::initialize and its brethren

evaleev:
(finalize, abort, fence, execute, default_execution_context) + updated tests/examples to use them

- These are top-level user runtime points dispatching to the default backends. It should be possible to support multiple backends co-existing in same program if needed, but some unresolved issues
- This also makes thread count handling uniform across all backends.


ttg/ttg/run.h
```diff
@@ -0,0 +1,61 @@
...
+  template <typename... RestOfArgs>
+  inline void initialize(int argc, char **argv, int num_threads, RestOfArgs &&...args) {
+    if (num_threads < 1) num_threads = detail::num_threads();
+    TTG_IMPL_NS::ttg_initialize(argc, argv, num_threads, std::forward<RestOfArgs>(args)...);
+  }
+
+  /// Finalizes the TTG runtime
+
+  /// This will possibly try to release as many resources as possible (some resources may only be released at
+  /// the conclusion of the program. Execution of TTG code is not possible after calling this.
```
devreal, line 32: Missing `)`

ttg/ttg/run.h
```diff
@@ -0,0 +1,61 @@
...
+  /// @internal ENABLE_WHEN_TTG_CAN_MULTIBACKEND To finalize the TTG runtime with multiple backends must call the
+  /// corresponding `ttg_finalize` functions explicitly.
+  inline void finalize() { TTG_IMPL_NS::ttg_finalize(); }
+
+  /// Aborts the TTG program using the default backend's `ttg_abort` method
+  inline void abort() { TTG_IMPL_NS::ttg_abort(); }
+
+  /// Accesses the default backend's default execution context
+
+  /// @note Dispatches to the `ttg_default_execution_context` method of the default bqckend
```
devreal, line 43:
```suggestion
  /// @note Dispatches to the `ttg_default_execution_context` method of the default backend
```

ttg/ttg/run.h
```diff
@@ -0,0 +1,61 @@
...
+  /// @return the default backend's default execution context
+  inline World default_execution_context() { return TTG_IMPL_NS::ttg_default_execution_context(); }
+
+  /// Starts the execution in the given execution context
+
+  /// @param world  an execution context associated with the default backend
+  /// @note Dispatches to the `ttg_execute` method of the default bqckend
+  inline void execute(World world = default_execution_context()) { TTG_IMPL_NS::ttg_execute(world); }
+
+  /// Returns when all tasks associated with the given execution context have finished on all ranks.
```
devreal, line 53: Should we add that a fence is a collective operations (like initialize and finalize, but unlike execute and abort)?

ttg/ttg/util/env.cpp
```diff
@@ -0,0 +1,33 @@
+//
+// Created by Eduard Valeyev on 11/6/21.
```
devreal, line 2: Do we have proper copyright headers?

ttg/ttg/util/env.h
```diff
@@ -0,0 +1,22 @@
...
+// Created by Eduard Valeyev on 11/5/21.
+//
+
+#ifndef TTG_UTIL_ENV_H
+#define TTG_UTIL_ENV_H
+
+namespace ttg {
+  namespace detail {
+
+    /// Determine the number of compute threads to use by TTG when not given to `ttg_initialize`
```
devreal, line 11:
```suggestion
    /// Determine the number of compute threads to use by TTG when not given to `ttg::initialize`
```
?

ttg/ttg/util/env.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <thread>
+
+#include <cstdlib>
+
+namespace ttg {
+  namespace detail {
+
+    int num_threads() {
+      std::size_t result = 0;
```
devreal, line 15: We may want to store the result in a static variable to avoid reparsing every time this function is called (maybe not done right now, but might be in a future use-case)

devreal (changes requested):

devreal:
@evaleev I really like this, it's been bothering me from the start :+1:

ttg/ttg/util/env.cpp
evaleev: currently `initialize` is only called once so this is not an issue ... plus the user may want to change the environment variable value programmatically, so I think we must query it every time.

ttg/ttg/util/env.h
evaleev: fixed

ttg/ttg/util/env.cpp
evaleev: not that I know of ... we should discuss

ttg/ttg/run.h
evaleev: fixed ... but isn't `execute` also collective?

ttg/ttg/run.h
evaleev: fixed

ttg/ttg/run.h
evaleev: fixed

README.md
```diff
@@ -98,7 +98,7 @@ auto make_sw1(const funcT& func, int block_size, const std::string &a, const std
 - Step 4 : Define the edges and verify that the graph is connected in the main program.
 
 ```cpp
-  ttg_initialize(argc, argv, -1);
+  initialize(argc, argv, -1);
```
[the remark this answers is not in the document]
evaleev: fixed .. the needs of the first-page dox  are still gaping :)

devreal:
Looks good. Just realized again: why do we have two different ways to query the default world? I see `ttg::get_default_world` and `ttg::default_execution_context`. Should we converge on one?

evaleev:
> Looks good. Just realized again: why do we have two different ways to query the default world? I see `ttg::get_default_world` and `ttg::default_execution_context`. Should we converge on one?

good catch, I think `default_execution_context` makes more sense

PR #198, opened 2021-11-18
PaRSEC: make sure mutable moved data copies are handled correctly

devreal:
No need to create a new copy if a mutable data was moved to send, so mark it as immutable. Also, fix some downstream bugs that remained undetected because of this.

I knew this code path was shaky, performance debugging the chain benchmark turned up this issue.


devreal:
Superseded by https://github.com/TESSEorg/ttg/pull/206, closing.

PR #206, opened 2022-01-11
PaRSEC: improve data copy handling

devreal:
This PR does 3 things:

1) Correctly mark mutable data copies as immutable if they are moved into send/bcast. They cannot be mutated after the call so they become read-only.
2) Consolidate the life-time management of data copies to only use the `reader` field in `parsec_data_copy_t`: there can be either a single writer or >=1 readers, enough information to release copies when they are not needed anymore. No need to use the `PARSEC_OBJECT_RETAIN/RELEASE` infrastructure.
3) Move the actual data object into the data copy to half the number of new/delete necessary. Instead of a custom delete mechanism we can use polymorphism and virtual destructors to facilitate proper destruction of data. Eventually we may want to use a free-list for data copies but I'm worried that we're exhausting memory by doing that. 

Changes 2 and 3 also allow us to get rid of the ttg_parsec.so library, which was only there for the C constructor of ttg_data_copy_t that's not needed anymore.


ttg/ttg/parsec/ttg.h
```diff
@@ -554,40 +532,54 @@ namespace ttg_parsec {
...
-          int32_t readers = parsec_atomic_fetch_dec_int32(&copy->readers);
-          if (1 == readers) {
-            copy->delete_fn(copy->device_private);
-            copy->device_private = NULL;
-          }
-        }
-      }
       if (NULL != copy->push_task) {
-        /* Release the task if it was deferrred */
+        /* Release the the deferred task.
```
bosilca, line 536: typo, double 'the'.

ttg/ttg/parsec/ttg.h
```diff
@@ -554,40 +532,54 @@ namespace ttg_parsec {
...
-        if (parsec_atomic_cas_ptr(&copy->push_task, push_task, nullptr)) {
-          parsec_ttg_task_base_t *deferred_op = (parsec_ttg_task_base_t *)copy->push_task;
-          assert(deferred_op->deferred_release);
-          deferred_op->deferred_release(deferred_op->tt_ptr, deferred_op);
+        copy->push_task = nullptr;
+        parsec_ttg_task_base_t *deferred_op = (parsec_ttg_task_base_t *)push_task;
+        assert(deferred_op->deferred_release);
+        deferred_op->deferred_release(deferred_op->tt_ptr, deferred_op);
+      } else {
+        if (copy->readers < 0) {
```
bosilca, line 546: How it is possible to have a negative number of readers ? Especially when the current task is supposed to hold at least one of the reader positions.

ttg/ttg/parsec/ttg.h
devreal: Fixed, thanks!

ttg/ttg/parsec/ttg.h
devreal: A negative value in readers marks the copy as being mutable. So the copy either belongs `readers > 0` (a bunch of readers) or `readers == INT32_MIN` (a single task mutating the copy).

devreal: Fix to check for `INT32_MIN` instead of smaller-than.

therault (approved):

PR #208, opened 2022-01-14
Start working on the front page of user documentation

therault:
TODO

- [ ] better example than Nth Fib?

  - ideas: binary search? mad-1d-project + integrate?

- [ ] User guide

  - [ ] rename to tutorial, and link to it from README (and vice versa)

  - [ ] use make_tt, remove tuples of terminals, use auto where appropriate

  - [ ] split the "diamond" example into 2, one with 2 separate TTs for B tasks, then demonstrate how parametrization allows to represent them as a single TT
  
- [x] `ttg::rank`: via #232

- [x] `ttg::once` and `TTG_ONCE`: : via #232

- [ ] `TTBase::invoke` to take pack of values instead of `std::tuple` 

- [ ] DAG pics: requires #222


README.md
```diff
@@ -28,18 +28,18 @@ This is the C++ API for the Template Task Graph (TTG) programming model for flow
     - `TTG_USE_MADNESS`: selects the MADNESS backend as the default (expert-use only).
 
     For simple programs it is sufficient to include the top-level header file:
-    ```cpp
-    #include <ttg.h>
-    ```
+```cpp
+#include <ttg.h>
+```
```
devreal, line 33: Somethign is wrong with the typesetting here:
Screenshot from 2022-03-04 16-44-35

devreal:
The main landing site still has `make_sw1` example, which is quite convoluted. Maybe we can replace that with a simpler one and link to the user guide?

CMakeLists.txt
```diff
@@ -39,6 +39,8 @@ set(CMAKE_INSTALL_CMAKEDIR "lib/cmake/ttg"
 ########################################
 option(TTG_PARSEC_USE_BOOST_SERIALIZATION "Whether to select Boost serialization methods in PaRSEC backend" ON)
 option(TTG_EXAMPLES "Whether to build examples" OFF)
+option(TTG_DOC_EXAMPLES "Whether to build the documentation examples" OFF)
```
evaleev, line 42: do we need separate control for building these examples?

devreal (approved):

PR #214, opened 2022-01-25
[docker] fix MPICH build issues + bump to clang 9

evaleev:


devreal:
@evaleev it seems like there is an unresolved cmake dependency to TBB in the new madness version?

evaleev:
CI ubuntu build issues will be fixed via #215

PR #217, opened 2022-02-08
Store input types in tuple

devreal:
Also, use existing `TT::ttT` base type definition instead of redefining the base type every time.

The use of type packs prevents us from using defaulted template arguments. I would like to use this change as a basis for future extensions, e.g., defaulted policies.

There is still an issue void input arguments, needs some more digging I guess...


devreal:
So, it turns out it's not that simple (why would it be). We can use `std::tuple<void>` as long as we don't call external functions with the `this` pointer. At that point the compiler tries to instantiate the void tuple, although it is unclear to me why that happens. Here is a reproducer:

```
#include <tuple>

void foo(void *);

template <typename TupleT>
class TT {
public:
    TT(){}

    void bar() {
    }

    void work() {
      bar();       // OK
      foo(this);   // NOT OK
    }
};

TT<std::tuple<void>> tt;
```
Unless I can figure out why and how to work around it, we have to stick with variadic templates. Bummer :/

devreal:
@evaleev the last commit adds support for multiple control inputs. I still need to fix `make_tt` and on the way probably will remove the type checking of the lambda input arguments to allow for generic lambdas.

devreal:
I added support for generic callables and for marking input edges as const using `ttg::make_const`, demonstrated here: https://github.com/TESSEorg/ttg/pull/217/files#diff-7864d573b344fb91a22dc0b343eae090d7182fa8cb9bfb3d4ea715994bbd935dR187

ttg/ttg/edge.h
```diff
@@ -138,6 +145,69 @@ namespace ttg {
...
+    // pure control edge should be usable to fire off a task
+    template <typename Key = keyT, typename Value = valueT>
+    std::enable_if_t<ttg::meta::is_all_void_v<Key, Value>> fire() const {
+      return m_edge.fire();
+    }
+  };
+
+  template<typename keyT, typename valueT>
+  auto make_const(Edge<keyT, valueT>& edge) {
+    return Edge<keyT, detail::const_wrap_t<valueT>>(edge);
```
evaleev, line 209: would `Edge<id, const Value>` be sufficient?

devreal: Yeah, I was thinking about that. Somewhere we check for decayed types in edges. If we allow `const Value` in edges then users may use that directly. The `const_wrap_t` is like a tag for the edge that is not supposed to be used by users directly.

devreal:
I fixed a couple examples that used `BlockMatrix<T>&&` as input arguments. This is no longer possible because we now derive the argument types from the edge types, not from the op (which may be generic). I don't think it fits into our model anyway.

It seem like some of the changes negatively impact performance for some examples so they timeout. I suspect it's the change from `Control` to `void` for control flow edges. I just can't make sense of it right now...

evaleev:
> I fixed a couple examples that used `BlockMatrix<T>&&` as input arguments. This is no longer possible because we now derive the argument types from the edge types, not from the op (which may be generic). I don't think it fits into our model anyway.
> 
> It seem like some of the changes negatively impact performance for some examples so they timeout. I suspect it's the change from `Control` to `void` for control flow edges. I just can't make sense of it right now...

Let's discuss on Friday. I'm not sure why rvalue ref does not fit our model ...

We should be able to simplify the effort on the user part and reduce fragility by detecting the right way to invoke generic callable, this will be unfortunately TMP-heavy ... but doable.

devreal:
@evaleev It seems that can get around explicitly casting to a reference type in `call_func` by moving the value:

Instead of
```C
    using func_args_t = ttg::meta::tuple_concat_t<std::tuple<const Key&>, input_values_tuple_type, output_edges_type>;
    func(std::forward<Key>(key),
         baseT::template get<S, std::tuple_element_t<S + 1, func_args_t>>(std::forward<Tuple>(args_tuple))..., out);
```

we'd do
```C
    func(std::forward<Key>(key), std::move(std::get<S>(args_tuple))..., out);
```
At least the unit tests and examples with move input seem to compile. Of course there might be more surprises...

And I need to patch up the handling of empty input reference types in the backends. That will fix the timeouts.

devreal:
Mhh, never mind the `std::move`. That won't work, our unit tests just didn't cover all cases (it breaks with non-const reference parameters). I'm not sure how the call signature could be detected without trying all combinations of rvalue and lvalue-reference. Here is a SO post doing that for a single parameter, doing that for more seems really quite messy: https://stackoverflow.com/a/30810231/16791592

devreal:
I cleared up more issues with `invoke()` and realized that there was an inconsistency between the parsec and madness backends in the way non-local `invoke()` was handled. Instead of adding code to handle edge cases, we now create a virtual control input that is triggered by `invoke()` just like any other input. Ideally we would have a base class that provides all of the public interface functions, which then dispatch to a single implementation in the TT. That would mean more exploding templatees though...

evaleev:
re: deciphering generic lambda's args:

- `f(auto)` binds to both `const&` and `&&`
- `f(auto&)` binds to only to `const&`
- `f(auto&&)` binds to both `const&` and `&&`
- `f(const auto&)` binds to both `const&` and `&&`

[see here](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:73,endLineNumber:6,positionColumn:73,positionLineNumber:6,selectionStartColumn:73,selectionStartLineNumber:6,startColumn:73,startLineNumber:6),source:'%23include+%3Ctype_traits%3E%0A%23include+%3Ctuple%3E%0A%0Aauto+func0+%3D+%5B%5D(auto+key,+auto+%26%26datum0,+auto+%26datum1,+const+auto+%26datum2,+auto+datum3,+auto%26+outs)+%7B%7D%3B%0Astatic_assert(std::is_invocable_v%3Cdecltype(func0),+int,+const+float%26,+const+float%26,+const+float%26,+const+float%26,+std::tuple%3C%3E+%26%3E)%3B%0Astatic_assert(std::is_invocable_v%3Cdecltype(func0),+int,+float%26%26,+float%26%26,+float%26%26,+float%26%26,+std::tuple%3C%3E+%26%3E)%3B%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50.46929125787411,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:gsnapshot,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+(trunk)+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),k:14.329923418094447,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,fontUsePx:'0',tree:'1',wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+(trunk)+(Compiler+%231)',t:'0')),k:35.200785324031436,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

Thus we can recommend for the user to pass all read-only data with complex type via `auto&` and consumed data with complex type via `auto&&`; to gain finer-grained control the types can be specified explicitly (e.g. `Matrix&&` or `const Matrix&`).

How we would pass an datum of type `T` to a task:
- if it is used by a single task (hence it may be more efficient to consume it rather than read it) we pass it via `T&&` or via `const T&` (if the task's callable cannot bind to `T&&`)
- if it is used by multiple tasks we make copies and pass them via `T&&` or via `const T&` (if the task's callable cannot bind to `T&&`)

devreal:
> f(auto) binds to both const& and &&

Does the detection detect if an `auto` parameter is mutated inside the lambda? In other words, would the detection disqualify `const&` in that case?

evaleev:
Is_invocable is evaluated in the unevaluated context so no. There is no way
without user assistance to detect such things ... The best c++ can do is to
produce a compilation error

evaleev:
@devreal we are more or less these with generic lambdas, I kept the ability to annotate edges via `make_const`, but in principle no longer needed ... just need testing+dox(the latter probably after merging into @therault's dox branch)

evaleev (approved):

evaleev:
this looks good ... the only issue is whether to remove `make_const` ... not leaning either way, but since the need for it can be eliminated by proper type specification (`auto&` vs `auto&&`) and removing it would trim some internal (`Edge` specialization, etc.) and user-facing code (make_const) might be something to consider.

devreal:
If we don't have any need for `make_const` I'm ok with removing it. We just need to make sure we properly document how to mark immutable inputs since that has significant repercussions on performance

PR #218, opened 2022-02-15
Implement Policies

devreal:
Policies are a mechanism to steer the execution of tasks through a single entity controlling different properties. The interface is extensible to allow for the addition of future properties without breaking backwards compatibility. If desired, providing a custom policy allows the compiler to fully inline all calls and thus avoid a bunch of indirect calls through `std::function`. This is useful for small tasks and allows us to reduce the overhead of querying these properties.

Currently, three properties are supported:

1) Process map (aka keymap): map a key identifying a task to a process for execution. Implemented using the `int procmap(const Key&)` member.
2) Priority map: map a key identifying a task to a priority. Implemented using the `int priomap(const Key&)` member.
3) Inline map: map a key identifying a task to a max depth for inlining. Implemented using the `int inlinemap(const Key&)` member. This has yet to be implemented in the backends.

Policies can be specified in two ways:

1) Using `make_policy()` to work with `make_tt`. Example:
```C
return ttg::make_tt([](){ /* task function */ }, edges(...), edges(...), 
                ttg::make_policy([](const Key& key){ return key % world.size(); }, 
                                 [](const Key& key){ return key.x; /* priority */}));
```
Functions are defaulted to make sure that all necessary properties are implemented. A function that is not provided can later be set dynamically on the tt (using `tt->set_inlinemap([](const Key&){ return 0; })` in the example above).

2) Using a custom policy class:

```C
struct MyPolicy : public ttg::TTPolicyBase {
  ttg::World world;
  MyPolicy(ttg::World world)
  : world(world), TTPolicyBase()
  { }
  int procmap(const Key& key) {
    return key % world.size();
  }
};

struct MyTT : ttg::TT<..., MyPolicy>
{
  ...
}
```
The `procmap` member function will shadow the `procmap` member of `ttg::TTPolicyBase`. Since the TT is templated on the policy, the `procmap` of `MyPolicy` is used. There is no need for virtual functions here. Similarly to the above, properties not specified in `MyPolicy` are defaulted to a `std::function` and can be set dynamically.

This PR is based on #217 to be able to default the policy type of TT so the diff is bigger than it will be if #217 gets merged.

### DIFF
--- a/examples/floyd-warshall/floyd_warshall.cc
+++ b/examples/floyd-warshall/floyd_warshall.cc
@@ -95,8 +95,7 @@ std::ostream& operator<<(std::ostream& s, const Control& ctl) {
 
 class Initiator : public TT<int, std::tuple<Out<Key, Control>, Out<Key, Control>, Out<Key, Control>, Out<Key, Control>>,
                             Initiator> {
-  using baseT =
-      TT<int, std::tuple<Out<Key, Control>, Out<Key, Control>, Out<Key, Control>, Out<Key, Control>>, Initiator>;
+  using baseT = typename Initiator::ttT;
 
  public:
   Initiator(const std::string& name) : baseT(name, {}, {"outA", "outB", "outC", "outD"}) {}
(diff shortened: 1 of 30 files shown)


devreal:
Superseded bu #226, closing

PR #219, opened 2022-02-17
Ashawini27/fix/add debug capabilities

Ashawini27:
Add debug capabilities to unit tests


devreal:
LGTM :+1:  It would be nice to have that functionality as part of a function/macro that we can drop into all our tests to enable debugging if `TTG_DEBUGGER` (or `TTG_ENABLE_DEBUGGER`?) is set.

README.md
```diff
@@ -14,6 +14,10 @@ This is the C++ API for the Template Task Graph (TTG) programming model for flow
 - To try out TTG in a Docker container, install Docker, then execute `bin/docker-build.sh` and follow instructions in `bin/docker.md`;
 - See [INSTALL.md](https://github.com/TESSEorg/ttg/blob/master/INSTALL.md) to learn how to build and install TTG.
 
+# Debugging TTG Program 
+- If xterm already installed then set "Environment Variable" to /TTG_DEBUGGER=gdb_xterm/lldb_xterm based on the type of compiler system hold to run debugger for parallelly executing processes
```
evaleev, line 18:
```suggestion
- If an X11 server is running (check if environment variable `DISPLAY` is set), then set environment variable `TTG_DEBUGGER` to {`gdb_xterm`,`lldb_xterm`} to launch {`gdb`,`lldb`} upon receiving a signal like `SIGSEGV` or `SIGABRT` (one `xterm` window per rank will be created);
- If an X11 server is not running the set `TTG_DEBUGGER` to empty value; upon receiving a signal the program will print instructions for how to attach a debugger to a running process from another terminal. 
```

README.md
```diff
@@ -1,4 +1,4 @@
-![Build Status](https://github.com/TESSEorg/ttg/workflows/CMake/badge.svg)
+![Build Status](https://github.com/TESSEorg/ttg/workflows/CMake/badge.sv)
```
evaleev, line 1: this change seems to be not intended, revert

evaleev (changes requested):

evaleev (approved):

PR #221, opened 2022-03-04
Implementation of MapReduce using TTG

pnookala:
This example counts the words in given files and prints the output to screen. Reduction is inefficient since it happens in a single process. This example works with the parsec backend, but there is an issue with the madness backend where it complains about unprocessed tasks, but the tasks actually get executed. This could be a problem with hashing std::string and needs to be fixed.


examples/mapreduce/mapreduce.cc
```diff
@@ -0,0 +1,270 @@
...
+#include <chrono>
+#include <filesystem>
+#include "ttg.h"
+
+#define BLOCK_SIZE 32
+
+using namespace ttg;
+
+template<typename T>
+using Key = std::pair<std::pair<std::string, T>, T>;
```
devreal, line 18: Do we need to send the filename as part of the key? Is it used after reading from the file?

examples/mapreduce/mapreduce.cc
```diff
@@ -0,0 +1,270 @@
...
+  chunk.erase(std::remove_if(chunk.begin(), chunk.end(),
+          []( auto const& c ) -> bool { return ispunct(c); } ), chunk.end());
+  std::istringstream ss(chunk);
+  std::string word;
+
+  while (ss >> word)
+  {
+    std::transform(word.begin(), word.end(), word.begin(), ::tolower);
+    //std::cout << "Mapped " << word << std::endl;
+    resultMap.insert(std::make_pair(word, 1));
```
devreal, line 121: What if the chunk contains the same word twice? And why use a multimap in the first place?

pnookala: I used multimap to be able to hold duplicate keys and have them sorted which makes counting easier in this example.

examples/mapreduce/mapreduce.cc
pnookala: This example counts words from multiple files. In order to keep the keys unique, I included filename in the key. We can probably come up with a different way of having unique keys as well without using the filename like a file ID for example.

PR #222, opened 2022-03-11
Modularize POTRF, make DPLASMA optional.

therault:
Make POTRF composable via make_ttg.

The goal is threefold:

  - remove the dependency on DPLASMA, at least for computation, if not for checking
  - have another example of graphs composition
  - prepare the fields to have a demonstration of speedup via graph composition


therault:
Superseded by PR #238

PR #223, opened 2022-03-15
make tuple of output terminals optional as an argument to TT's callable and `DerivedTT::op`

evaleev:
the plan is to have a thread-local pointer to `array<TerminalBase*>`, set by the backend when invoking the function. `ttg::send<I>(key, data...)` would then grab the corresponding `TerminalBase*`, cast it to the concrete type and push the data into it.

There will be no overhead for the current code that passes around the terminal tuple explicitly. Passing the tuple implicitly will make the code more fragile but can use RTTI (when in debug mode) to type match between the terminal type and the datum at runtime.

### DIFF
--- a/ttg/ttg/func.h
+++ b/ttg/ttg/func.h
@@ -40,14 +40,14 @@ namespace ttg {
     };
 
     template<typename keyT, typename valueT>
-    inline auto get_dynamic_terminal(size_t i, const char* func) {
+    inline auto get_out_terminal(size_t i, const char* func) {
 #ifndef NDEBUG
       auto *base_terminal_ptr = TTBase::get_outputs_tls_ptr()->at(i);
       auto *terminal_ptr = dynamic_cast<Out<std::decay_t<keyT>, std::decay_t<valueT>> *>(base_terminal_ptr);
       if (terminal_ptr == nullptr) {
         std::stringstream ss;
         ss << func << ": invalid type of ith output terminal, most likely due to mismatch between its type "
-                      "and the type of key/value; make sure that the arguments to send() match the types encoded in the output "
+                      "and the type of key/value; make sure that the arguments to " << func << "() match the types encoded in the output "
                       "terminals, or pass the output terminal tuple to the task function explicitly";
         throw std::runtime_error(ss.str());
       }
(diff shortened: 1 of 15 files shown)


ttg/ttg/func.h
```diff
@@ -95,162 +92,252 @@ namespace ttg {
...
+  std::enable_if_t<!meta::is_void_v<keyT> && is_output_terminal_v<output_terminalT>, void> set_size(
+      const keyT &key, const std::size_t size, output_terminalT &t) {
     t.set_size(key, size);
   }
 
   template <size_t i, typename keyT, typename... output_terminalsT>
-  std::enable_if_t<!meta::is_void_v<keyT>,void>
-  set_size(const keyT &key, const std::size_t size, std::tuple<output_terminalsT...> &t) {
+  std::enable_if_t<!meta::is_void_v<keyT> && (is_output_terminal_v<output_terminalsT> && ...), void> set_size(
+      const keyT &key, const std::size_t size, std::tuple<output_terminalsT...> &t) {
```
devreal, line 303: God, rly? Put the function name at the end of the line...?

Anyway, why not be more specific on the output terminal types to use overload resolution without the manual constraints here by doing something like:

```C
template <size_t i, typename keyT, typename... output_terminal_valuesT>
  void set_size(
      const keyT &key, const std::size_t size, std::tuple<ttg::Out<keyT, output_terminal_valuesT>...> &t) {
```
We should be able to do that for all functions here except for the broadcast, where we can have several key types.

ttg/ttg/make_tt.h
```diff
@@ -44,15 +46,32 @@ class CallableWrapTT
 
   template <typename Key, typename Tuple>
   void call_func(Key &&key, Tuple &&args, output_terminalsT &out) {
-    func(std::forward<Key>(key), std::forward<Tuple>(args), out);
+    if constexpr (funcT_receives_outterm_tuple)
+      func(std::forward<Key>(key), std::forward<Tuple>(args), out);
+    else {
+      this->set_outputs_tls_ptr();
```
devreal, line 52: Should this be moved into the backend TT (or some sort of wrapper) so that it can be used by applications with custom TT classes?

devreal (changes requested):
I like it :+1: Just two comments

ttg/ttg/func.h
evaleev: yes, correct, I did not complete the conversion

PR #238, opened 2022-09-12
TTG composition, example of POINV, and a few fixes to parsec profiling

therault:
This supersedes the PR #222.

- showcase TTG composition via the example of POINV
- enable a new scheduling policy in the parsec backend: defer-writer, where writer tasks are delayed until there is no reader task (a safeguard prevents live-locks by allowing a given task to be deferred indefinitely)
- expose profiling and DAG of tasks outputing options to the user, enable these options with the parsec backend


devreal:
@evaleev I added a test to check that the backend receives the right type (in terms of constness) from the auto detection of types: https://github.com/TESSEorg/ttg/pull/238/files#diff-7864d573b344fb91a22dc0b343eae090d7182fa8cb9bfb3d4ea715994bbd935dR314

The problem is that the `static_assert` in the lambda works (the type inside the lambda is correctly `const`) but the backend doesn't detect it as const:  

```
26: /home/runner/work/ttg/ttg/tests/unit/tt.cc: In function ‘void ____C_A_T_C_H____T_E_S_T____0()’:
26: /home/runner/work/ttg/ttg/tests/unit/tt.cc:314:28: error: static assertion failed
26:   314 |         static_assert(std::is_const_v<std::tuple_element_t<0, tt_t::input_args_type>>);
26:       |                       ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is bad because it forces us to create copies where we don't need them in the PaRSEC backend.

I wonder if we actually can get this to work properly. Swapping the order in which we test the types of course doesn't help because then `auto&&` becomes `const T&` and other asserts fail. I'm not entirely sure how the current implementation works but it seems that trying to bind `T&&` to `const T&` works and then of course inside the lambda it's correctly detected as `const T&` because that's what the parameter type enforces... Any ideas?

devreal:
I believe the detection of constness of arguments for `auto` is impossible: both `auto&&` and `const auto&` can bind to anything, surprisingly even `T&&`:

```C
    auto movefn = [](auto&& x){ return x; };
    auto constfn = [](const auto& x){ return x; };

    int x;
    movefn(static_cast<const int&>(x));
    constfn(static_cast<int&&>(x));
    constfn(std::move(x));
```
All compiles well in godbolt.

Not sure where to go from here but we cannot advertise this in the tutorial if it doesn't work as expected. The most important aspect of constness in TTG is the ability to elide copies and we cannot do that correctly. We can skim over that problem for the tutorial but eventually we need to find a way to properly advertise generic const arguments.

evaleev:
@devreal "const" (read-only) generic parameters should be denoted `auto&`, consumable generic parameters should be denoted `auto&&`. I know that it is counterintuitive that using `const auto&` is also allowed for the consumable parameters, that's why the dox only mention `auto&` and `auto&&`: https://github.com/TESSEorg/ttg/blob/master/ttg/ttg/make_tt.h#L406

evaleev:
added a warning to the `make_tt()` dox + improved/annotated test explaining why `const auto&` is to be avoided

ttg/ttg/parsec/ttg.h
```diff
@@ -606,7 +606,7 @@ namespace ttg_parsec {
         }
 
         // We store the hash of the key and the address where it can be found in locals considered as a scratchpad
-        uint64_t hv = ttg::hash<decltype(key)>{}(key);
+        uint64_t hv = ttg::hash<std::decay_t<decltype(key)>>{}(key);
```
evaleev, line 609: @devreal seems like we hashed refs to keys rather than keys themselves here ... not sure how this worked

evaleev (approved):

PR #240, opened 2022-09-13
HOTFIX: make_tt without explicit out parameter does not manage well inline/recursive calling of function

therault:
Some backends (MADNESS) allow to recursively call the function during… the call to send, this the TLS for the outputs need to be stored and restored before/after the function call, or we may work with the wrong outputs


therault:
We (me and @devreal) noticed an erroneous behavior with MADNESS when using implicit out parameter in `make_tt`: the second `send` call within a function would use the wrong out. We believe this is because MADNESS allows the `send` to directly call the function, having a side effect on the `outputs_tls`, which needs to be restored after the function is called.

evaleev (approved):

PR #241, opened 2022-09-22
Adapt the github workflow to run on nektos/act

devreal:
See https://github.com/nektos/act

Not sure we want to merge it. I'm still seeing the following errors for the unittests:

```
| 25: terminate called after throwing an instance of 'std::runtime_error'
| 25:   what():  locale::facet::_S_create_c_locale name not valid
```

That only occurs on the core-unittests though, all other tests succeed on my machine.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -38,11 +38,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Create Build Environment
-      # Some projects don't allow in-source building, so create a separate build directory
-      # We'll use this as our working directory for all subsequent commands
-      run: cmake -E make_directory ${{github.workspace}}/build
-
     - name: Install prerequisite MacOS packages
       if: ${{ matrix.os == 'macos-latest' }}
       run: brew install ninja gcc@10 boost eigen open-mpi bison ccache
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -41,7 +44,16 @@ jobs:
     - name: Create Build Environment
       # Some projects don't allow in-source building, so create a separate build directory
       # We'll use this as our working directory for all subsequent commands
-      run: cmake -E make_directory ${{github.workspace}}/build
+      run: |
+          echo "tzdata tzdata/Areas select America" > ubuntu-preseed.txt
```
evaleev, line 48: this should go into ubuntu package install section, messes up the macOS builds

.github/workflows/cmake.yml
```diff
@@ -41,7 +44,16 @@ jobs:
...
+          echo "tzdata tzdata/Zones/Etc select UTC" >> ubuntu-preseed.txt
+          echo "tzdata tzdata/Zones/America select New_York" >> ubuntu-preseed.txt
+          debconf-set-selections ubuntu-preseed.txt
```
evaleev, line 51: sounds like this also needs `sudo`?

evaleev (changes requested):

.github/workflows/cmake.yml
devreal: Removed, didn't work and should instead be fixed by removing the `setlocale` call in the unit test.

.github/workflows/cmake.yml
devreal: I had to move the make directory creation after the package installation to make sure we have CMake available.

evaleev (approved):

PR #246, opened 2022-10-14
Mock of ttg::make_device_tt and views

devreal:
This is the beginning of the device journey. A mock that should at least compile but doesn't do anything useful yet.

@therault and I came up with the API used in the example below. This mock should serve as a starting point to play around with the API. 

At the core, we introduce `ttg::make_device_tt` that takes three callables:
1) Prologue: Creating views of data on the host, which are either reused (based on what we're tracking in PaRSEC) or created new on the device.
2) Kernel: Launching kernels on the device (and potentially enqueuing transfers back out of the device). The views created in the prologue are passed to this callable and allow querying both host side objects and device pointers.
3) Epilogue: called after all device kernels completed, used to send data to next tasks.

A second version of `ttg::make_device_tt` takes four callables where the first one is the host tasklet to allow the backend to schedule it either way.

The next steps will be to implement the data transfer functionality and kernel launch in the PaRSEC backend. Feedback on the design is welcome :) 

```C
auto make_gemm(ttg::Edge<Key3, MatrixTile<T>>& A,
               ttg::Edge<Key3, MatrixTile<T>>& B,
               ttg::Edge<Key2, MatrixTile<T>>& output_result)
{

  auto f_cpu = [=](const Key3& key,
                   const MatrixTile<T>&  A,  
                   const MatrixTile<T>&  B,  
                   MatrixTile<T>&  C,  
                   std::tuple<ttg::Out<Key2, MatrixTile<T>>,
                              ttg::Out<Key3, MatrixTile<T>>>& out)
  {
    int m = std::get<0>(key);
    int n = std::get<1>(key);
    int k = std::get<2>(key);

    if(k == 0) {
      dlprng(C.data(), 1789, A.mb()*B.nb());
    }   
    dgemm(A.data(), A.mb(), A.nb(),
          B.data(), B.mb(), B.nb(),
          1.0,
          C.data(), C.nb(), C.nb());

    if( k == KT-1 || C.data()[0] < 1e-9 ) { 
      ttg::send<0>(Key2{m, n}, std::move(C));
    } else {
      ttg::send<1>(Key3{m, n, k+1}, std::move(C));
    }   
  };  
  
  auto f_gpu_host_views = [=](const Key3& key,
                              const Matrixtile<T>& A,
                              const MatrixTile<T>& B,
                              MatrixTile<T>* C)  
  {
    ttg::View<const MatrixTile<T>, const T> dev_A = ttg::make_view( A, ttg::ViewSpan(A.data(), A.nb_data()) );
    ttg::View<const MatrixTile<T>, const T> dev_B = ttg::make_view( B, ttg::ViewSpan(B.data(), B.nb_data()) );
    ttg::View<MatrixTile<T>, T> dev_C;
    ttg::View<T, T> dev_tmp;
    T *host_tmp = new(T);
    dev_tmp = ttg::make_view( *host_tmp, ttg::ViewSpan(host_tmp, 1, ttg::ViewScope::SyncOut) ); // dev_tmp is a promise of 1 T on the device, associated with host_tmp
    
    int k = std::get<2>(key);
    if(0 == k) {
      // ViewScope::Allocate tells the runtime system that the device view needs to be allocated but doesn't need to be
      // initialized with C.data(). However, C.data() is still associated with the device memory, so if the
      // runtime system evicts that data from the device, it will be first copied back into C.data().
      dev_C = ttg::make_view( C, ttg::ViewSpan(C.data(), C.nb_data(), ttg::ViewScope::Allocate) );
    } else {
      // The default is ViewScope::SyncIn, i.e., data is copied from the host to the device if necessary
      dev_C = ttg::make_view( C, ttg::ViewSpan(C.data(), C.data_size()) );
    }   
    
    return std::make_tuple(dev_A, dev_B, dev_C, dev_tmp);
  }

  auto f_gpu_kernel = [=](const Key3& key,
                          ttg::View<const Matrixtile<T>, const T>& dev_A,
                          ttg::View<const Matrixtile<T>, const T>& dev_B,
                          ttg::View<Matrixtile<T>, T>& dev_C,
                          ttg::View<T, T>& dev_tmp)
  {
    int k = std::get<2>(key);

    const MatrixTile<T>&  A = dev_A.get_host_object();
    const MatrixTile<T>&  B = dev_B.get_host_object();
    MatrixTile<T>&  C = dev_C.get_host_object();
    T& host_tmp = dev_tmp.get_host_object();
    auto beta = 1.0;
    if(k == 0) {
        cublasDplrng(dev_C.template get_device_ptr<0>(), C.mb(), C.nb());
    }

    cublasDgemm(dev_A.template get_device_ptr<0>(), A.mb(), A.nb(),
                dev_B.template get_device_ptr<0>(), B.mb(), B.nb(),
                beta,
                dev_C.template get_device_ptr<0>(), C.mb(), C.nb());
  };

  auto f_gpu_output_flows = [=](const Key3& key,
                                const MatrixTile<T>&  A,
                                const MatrixTile<T>&  B,
                                MatrixTile<T>&  C,
                                T& host_tmp,
                                std::tuple<ttg::Out<Key2, MatrixTile<T>>,
                                           ttg::Out<Key3, MatrixTile<T>>>& out)
  {
    int m = std::get<0>(key);
    int n = std::get<1>(key);
    int k = std::get<2>(key);

    if( k == KT-1 || host_tmp < 1e-9 ) {
      ttg::send<0>(Key2{m, n}, std::move(C));
    } else {
      ttg::send<1>(Key3{m, n, k+1}, std::move(C));
    }
    delete &host_tmp;
  };

  /* If we only have GPU */
  auto gemm_tt = ttg::make_device_tt(f_gpu_host_views, f_gpu_kernel, f_gpu_output_flows, ttg::DEV_CUDA,
                                     ttg::edges(A, B), ttg::edges(output_result, C),
                                     "GEMM", {"A", "B"}, {"output_result", "C"});

  /* Alternative: to get both type of tasklets: */
  auto gemm_tt = ttg::make_device_tt(f_cpu, f_gpu_host_views, f_gpu_kernel, f_gpu_output_flows, ttg::DEV_CUDA,
                                     ttg::edges(A, B), ttg::edges(output_result, C),
                                     "GEMM", {"A", "B"}, {"output_result", "C"});

  return gemm_tt;
}
```


devreal:
TODO:

- Provide synchronous views that always sync back to the host before calling the host output function

robertjharrison:
This seems not consistent with the design agreed upon at the Roanoke meeting.   It also seems to leave the user still responsible for managing data motion whereas I understood that PaRSEC would be managing the motion of data in DCs.  It also elevates the concepts of views into being required instead of just a convenience.  Overall, I find it cluttered with too many concepts for users to make sense of.  The point of the Roanoke design was that writing a TTG gpu kernel should be no harder than writing some standard CUDA kernel and ideally easier since most data motion would be handled by PaRSEC.

therault:
> This seems not consistent with the design agreed upon at the Roanoke meeting. It also seems to leave the user still responsible for managing data motion whereas I understood that PaRSEC would be managing the motion of data in DCs. It also elevates the concepts of views into being required instead of just a convenience. Overall, I find it cluttered with too many concepts for users to make sense of. The point of the Roanoke design was that writing a TTG gpu kernel should be no harder than writing some standard CUDA kernel and ideally easier since most data motion would be handled by PaRSEC.

We agree on this: we do not intend the user to do data movement. The example tried to show all features / possibilities and it includes one `cudaMemcpyAsync` to show how they could speedup data-dependent decisions / ensure minimal data movement by sending back into temporary memory the minimal data they need. But the vast majority of inputs and ouputs should be managed by the runtime system.

A maybe cleaner way for the user to achieve the same thing would be to declare the tmp as an input, and specify in the view creation that this data needs to be synchronized on the host when the kernel is done.

For views, we only envisioned cases where the user provides a view for all inputs, but we are not opposed to a policy that says 'if the user does not define a view for that input -- or does not provide a building view function at all --, the runtime will assume trivially copyable inputs and use this as the view of the object.

robertjharrison:
A virtue of this design (that I still think is way to complex and too much
burden over just writing a native CUDA kernel) is at least that it seems
backend neutral since allocation and data motion seem to be explicit and
user managed.

But where then is parsec helping with automating and optimizing data motion
and memory allocation?  What have I missed?

Also, is it possible to have multiple kernels and steps for data transfer
within 1 task?  The previous continuation-like design permitted a more
nuanced approach (e.g., copying additional data back or invoking additional
kernels based on runtime values).

evaleev:
This is the right design, Robert. Your concerns really reflect difficulty
of dealing with heterogeneous memory space and asynchrony. Experts-first
design ensures this will be useful, then can be simplified for trivial
types and for archs with unified memory space. But must support
heterogeneous space and nontrivial data types first.

Best discussed tomorrow during the weekly meeting.

robertjharrison:
Ed, I of course trust your judgement but I guess we differ a little in
emphasis.

I wish we as a project had adopted the practice of  actual documented
designs rather than the usual software soup ... Since I've been AWOL so
long I have no clue about how the design meshes with use cases and in this
I mirror potential external users.

I will be on the call tomorrow.

evaleev:
Robert: agree, having complete documented design description+rationale first would be ideal. Due to limited f2f time in Knoxville + limited amount of designing possible without compiler's help we decided to compile a mock to identify any missing gaps in design. Using complete example allowed to identify concept gaps and get a feel for design. But we did not focus on updating the design docs. I'll try to summarize our current understanding of the design space as I see it, it can be the basis for dumping it as a rationale for our design into design doc / paper.

---------

### problem statement
TTG must be able to execute general user-defined graphs on machines with heterogeneous execution and address spaces, e.g., using multiple processes each having multiple CPU threads + device streams, with each thread/stream preferring or limited to a specific address range.

### key concerns
The key issues are how to manage:
- the asynchrony of the device programming models, and
- the heterogeneity of the address space.

There are multiple "solutions" to each issue, hence there are many possible designs. I'll discuss each issue first, then outline the aggregate designs we are pursuing.

#### Memory:
- *Unified (UM)*: where available, use single address space (unified memory visible to both host and device executors; it is also possible to use pinned host memory for device calls)
  - pro: simplifies memory management by removing the capacity limitation
  - con: still needs user cooperation: all compute data must be allocated on UM heap, this impacts the design of
  user data types, e.g. making them allocator aware, etc.
  - con: the user will likely needs to use pooled memory management for efficiency reasons (e.g., TiledArray uses Umpire)
  - con: still necessary to provide hints to the kernel driver managing UM.
  - con: reasoning about UM driver performance is difficult, its details are opaque and platform dependent.
- *Native (NM)*: using "native" device memory.
  - pro: simpler performance model due to greatest amount of control (by runtime) over execution
  - pro: can work with stack-capable data types
  - con: The amount is limited, hence this memory must be explicitly managed (akin to how a cache is managed).

#### Asynchrony
- *Stages*: decompose tasks into stages, with runtime-managed periods between stages for managing asynchrony of the actions scheduled by each stage
  - pro: most explicit, easier to reason about
  - con: most verbose; device-capable tasks look very different from host tasks
- *"Threads"*: use conventional mechanisms (threads for current backends, coroutines/fibers usable).
  - pro: least host/device dichotomy
    - threads: tasks are ordinary functions
    - fibers/coroutines: task functions "look" like ordinary functions (and can be made almost like normal functions using macros)
  - con: potentially lower performance
    - threads: due to the need to context switch to "yield" to other tasks
    - fibers/coroutines: implementation details are more complex and usually involve heap allocation (C++20 coro)
  - con: runtime is still responsible for managing the executor space heterogeneity (control where to launch a task) and asynchrony (events/host callbacks).

#### other considerations

- it's not possible to manage memory from the device code, hence all program logic, including _device-capable_ tasks, must execute on host executors. In principle if we restricted ourselves to a single-source language (SYLC-extended C++) we could write device capable tasks directly as device code, but current language limitations mandate wrapping everything into host code.

#### Current designs
- *UM+threads*: use UM for memory management + threads for asynchrony
- *NM+stages*: use Parsec's native memory pool manager + stage-decomposed tasks

evaleev:
I am looking at C++ coroutines as of this morning, this looks like a practical way to write device-capable tasks, though not sure how available C++20 is on the big iron right now.

devreal:
This has been superseded by the Buffers concept.

PR #247, opened 2022-10-21
ciao dplasma

evaleev:


devreal (approved):
Too many whitespace changes, cannot review :/ Removing dplasma seems good though

evaleev:
@devreal thx, amended the formatting changes

ttg/ttg/util/multiindex.h
```diff
@@ -0,0 +1,78 @@
...
+// Created by Eduard Valeyev on 10/21/22.
+//
+
+#ifndef TTG_UTIL_MULTIINDEX_H
+#define TTG_UTIL_MULTIINDEX_H
+
+namespace ttg {
+
+  template <std::size_t Rank>
+  struct MultiIndex {
```
devreal, line 11: Should we make the index type a template argument with a default to `int` (should be sufficient for most applications). Right now, this PR would double the size of a broadcasts in POTRF because we're using `long` instead of `int`. It's unlikely that we need 64bit for each index.

PR #252, opened 2023-03-30
TTG over parsec master

therault:
This is a work in progress, about 90% achieved, and working version.

PaRSEC still needs to change, to enable a different lazy initialization of its communication engine.
The current lazy initialization disables tag registration, and the TTG DSL requires the capability to register tags before the communication system is running. The communication system should lazily initialize costly resources (MPI), but initialize the internal resources (array of callbacks, etc...) eagerly to allow tag registration to work at any time. The current workaround registers registration and de-registration callbacks that are called when the comm engine is up or gets down, but this is too convoluted.

Rest of the features should work and are implemented as intended (including profiling in distributed).


devreal (approved):

devreal:
Thanks @therault!

PR #255, opened 2023-04-21
fixed issues for handling files obtained for multiple ranks

Ashawini27:


evaleev:
this requires the user to pass the entire filename of PBT file for 1 rank, rather than prefix, which I think is worse requiring to pass the file _prefix_.

PR #256, opened 2023-05-26
PaRSEC TTG: Offload stream reductions

devreal:
When we encounter a streaming terminal we schedule a dedicated
reducer task. Subsequent inputs are put into a LIFO. Once the
reducer task runs it will process all available inputs for that
terminal but will not block waiting for all inputs. In the
worst case, we create N tasks (one for each input), likely less.

Currently we use the priority of the task we stream into but this should be revisited, maybe a higher priority could be useful.


devreal:
@therault the tests are failing because the madness tag is unavailable. Do you have a tag to use instead?

therault:
31b2470ca722a6a2d84d4de08d32fb72ae8fdeda

https://github.com/therault/madness/commit/31b2470ca722a6a2d84d4de08d32fb72ae8fdeda

PR #257, opened 2023-06-06
PaRSEC: replace thread-local variable with parsec_my_execution_stream()

devreal:
The thread-local variable parsec_ttg_es was a hack that's not needed anymore, and was buggy.


devreal:
Good catch! Removed in the last push

PR #258, opened 2023-06-09
Traverse all TTs even if an issue is found

devreal:
Not traversing all TTs leads to nasty deadlocks, e.g., the graph will execute but communication will hang because not all tasks were made executable.

There are valid cases where not all inputs are connected and the user explicitly sets these inputs, e.g., using `invoke`. We should support that and not silently deadlock and make everyone's life miserable.

### DIFF
--- a/ttg/ttg/traverse.h
+++ b/ttg/ttg/traverse.h
@@ -62,7 +62,7 @@ namespace ttg {
               std::cout << "ttg::Traverse: got a null predecessor!\n";
               status = false;
             } else {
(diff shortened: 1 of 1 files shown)


therault (approved):
`&=` uses the bitwise `&` which is not short-circuiting the call to the right-hand expression. This works, even if the implict casting `bool -> integer -> bool` is a bit confusing... The alternative (`bool rc = traverse(); status = status && rc;`) is more verbose.

LGTM.

PR #259, opened 2023-06-09
Cast keys passed to invoke() to known key types

devreal:
Otherwise the user controls the key type sent in messages to remote owners of tasks.

### DIFF
--- a/ttg/ttg/madness/ttg.h
+++ b/ttg/ttg/madness/ttg.h
@@ -187,6 +187,7 @@ namespace ttg_madness {
 
    public:
     using ttT = TT;
+    using key_type = keyT;
     /// preconditions
     static_assert((ttg::meta::none_has_reference_v<input_valueTs>), "input_valueTs cannot contain reference types");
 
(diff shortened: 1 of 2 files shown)


therault (approved):

bosilca:
It is good to have things done automatically, but this does not inform the developer that we are potentially truncating things on his behalf. Can we have a warning message hinting at the type conversion ?

therault:
My understanding is that if the type passed to invoke() is not castable
into the key type defined in the TT creation, the compiler will complain
automatically.
In the example that generated the issue ( `using keyT = int; ... for(size_t
i = 0; i < 10; i++) tt->invoke(i, ...); } ` ) I assume such cast will
complain, because one cannot always cast a size_t into an int. Am I right
about the compiler complaining (at least with a warning)?

devreal:
It's not necessarily a truncation, might as well be a widening. We could add a one-time warning that we're doing the conversion. If the conversion is not possible (mismatching types) or the compiler warns about trunction it will bark. It does not do that by default, it seems

bosilca:
I'm not concerned that much about widening, or changing type that have the same expressivity range. I'm concerned that the user who wrote a code similar to yours will indeed expect the key to be a size_t, but internally we reduce it to an int without feedback.

devreal:
Yes, but the user expressed that the keys should be `int` when they defined the Edge. So if they provide a key/value larger than int they break their own contract. I'm more concerned about the key being `size_t` and the user passing literal integers (which by definition are `int`). The conversion is harmless there.

Alternative: error out if the types don't match but that seems less user-friendly.

therault:
Yeah, I think what we want is

if the user writes `ttg::Edge<int, ...> F2F; auto f = ttg::make_tt(
[](const int &a, ...) {...}, ttg::edges(F2F), ...); for(size_t i = 0; i <
3; i++) f->invoke(i, ...)` we complain / fail compiling, because they
defined `f` with an int and used it with a size_t

if the user writes `ttg::Edge<size_t, ...> F2F; auto f = ttg::make_tt(
[](const size_t &a, ...) { ...}, ttg::edges(F2F), ...); for(int i = 0; i <
3; i++) f->invoke(i, ...);` we silently upgrade i into a size_t in `invoke`.

As far as I understand, that's what this PR does: the added explicit cast
will succeed when we upgrade the input, but fail when we try to downgrade
it. The error (if there is one) will appear at the invoke() call, which is
where the user wrote something wrong, as @Joseph Schuchart
***@***.***> noted.

( and if the user writes `ttg::Edge<size_t, ...> F2F; auto f =
ttg::make_tt( [](const int &a, ...) {...}, ttg::edges(F2f), ...);` we
complain / fail to compile because the types of the input Edge and the type
of the lambda are not compatible. This has always been wrong and detected
by the compiler, as far as I know. )

PR #260, opened 2023-06-22
Synchronize on make_graph_executable

devreal:
PaRSEC needs to make sure that no AMs are delivered before the callbacks are registered, which happens in make_graph_executable. So we add a hook that becomes a barrier for PaRSEC and is empty for madness.

@Ashawini27 this should fix your hangs. Please give it a try and let me know if not :)


therault (approved):
LGTM

Issue #274, opened 2024-02-28
boost serialization: avoid virtual function calls


evaleev:
the optimized Boost archive provided by TTG still uses Boost infrastructure (`boost::iostreams::basic_array_sink`) for serialization to/from in-memory buffer. Unfortunately this Boost archive boilerplate (`boost::archive::binary_oarchive_impl` and it bases) performs serialization to byte stream via [std::basic_streambuf::xsputn](https://en.cppreference.com/w/cpp/io/basic_streambuf/sputn) which is virtual. Lack of compile-time inlining (presumably LTO would help) is an unnecessary burden.

Ideally Boost serialization can be inlined as much as possible without the need for LTO.

PR #275, opened 2024-02-28
Boost serialization: avoid virtual function calls

evaleev:
"optimized" archive provided by TTG for Boost serialization into in-memory buffers uses Boost infrastructure (`boost::archive::binary_{i,o}archive_impl`) that makes virtual function calls that cannot be inlined at compile time (see #274 ). This addresses the issue.


devreal (approved):
Thanks, these are good changes :+1:

PR #278, opened 2024-04-01
Use the correct device index in taskpool devices_index_mask

devreal:
We should use the linear index [0..nb_devices-1] instead of device->device_index.


ttg/ttg/parsec/ttg.h
```diff
@@ -339,7 +339,7 @@ namespace ttg_parsec {
       for(int i = 0; i < (int)parsec_nb_devices; i++) {
           parsec_device_module_t *device = parsec_mca_device_get(i);
           if( NULL == device ) continue;
```
therault, line 341: This is interesting... I'm not saying this is wrong, but I was under the impression that 
``` i == device->device_index```
should be a true invariant... Is there something more fishy? Or my understanding is wrong?

I will check what happens in `parsec_device_cuda_module_init()`

devreal:
Not needed, closing

PR #279, opened 2024-04-03
SPMM: add Boost::graph as dependency

devreal:
Brought up by @therault


therault (approved):
lgtm

cmake/modules/FindOrFetchBoost.cmake
```diff
@@ -17,12 +17,14 @@ set(required_components
         headers
         callable_traits
 )
+if ()
```
evaleev, line 20: ???

evaleev (approved):
Looks good, glad it was "this" simple

PR #280, opened 2024-04-04
Pass argc/argv to allocator_init

devreal:
They are needed to initialize madness and thus the TA allocators.


devreal:
@therault @evaleev please review and merge

evaleev (approved):

PR #281, opened 2024-04-04
Change guards for CUDA/HIP/L0 to TTG_ENABLE_*

devreal:
We can detect both HIP and CUDA on some machines but may only build one at the time.


therault (approved):
LGTM

PR #282, opened 2024-04-06
update `README` with device example

evaleev:
using data-dependent fibonacci example (find `max(F_n) < N`) as the motivating dataflow example


devreal (approved):
Thanks @evaleev! Changes look good. Technically, we could use move-only types in broadcasts but I'm fine with the stricter constraint. We can relax that later if we have to.

devreal:
This looks good but should we add a section on environment variables? I guess the only relevant one right now is `TTG_NUM_THREADS`.

evaleev:
@devreal I made some changes to the device example discussion, would appreciate your proof ... [here](https://github.com/TESSEorg/ttg/blob/master/README.md)

PR #283, opened 2024-05-22
Add device hints and return to PaRSEC main repo

devreal:
This PR does two things that got intermingled:

1) Add a mapper to device-enabled TTs to hint at which device to execute on. This almost doubled performance on Frontier.
Example hinting at keeping all tiles of a row on the same device (see potrf.h for actual use):
```
tt->set_devicemapper([](const Key& key){ return (key[0] / A.P()) % ttg::device::num_devices(); });
```

2) Return us to PaRSEC master after @abouteiller and @therault massaged PaRSEC long enough to make it fit our needs. Thanks! We now collect the input  data in the evaluate function so that device selection can happen before the main task callback executes.


ttg/ttg/parsec/device.h
```diff
@@ -35,6 +36,12 @@ namespace ttg_parsec {
     }
   } // namespace detail
 
+
+  inline
+  int num_devices() {
+    return parsec_nb_devices - detail::first_device_id;
```
abouteiller, line 42: We now have `parsec_context_query(ctx, PARSEC_CONTEXT_QUERY_DEVICES, PARSEC_DEV_CUDA)` that can be used to count devices without having to know about the first device id.

ttg/ttg/parsec/ttg.h
```diff
@@ -765,7 +766,7 @@ namespace ttg_parsec {
         parsec_ttg_task_t<TT> *me = (parsec_ttg_task_t<TT> *)parsec_task;
         return me->template invoke_op<ttg::ExecutionSpace::CUDA>();
       } else {
-        throw std::runtime_error("PaRSEC CUDA hook invoked on a TT that does not support CUDA operations!");
+        return PARSEC_HOOK_RETURN_NEXT;
```
abouteiller, line 769: It is now incorrect to return NEXT from a `hook`, it can be returned only during `evaluate`. This will cause parsec to invoke a fatal error. You shall return PARSEC_HOOK_RETURN_ERROR instead, but. that will also cause a fatal error, as it is too late in the game to change your mind (the data has been positioned on that device, and the owner device has been set).

ttg/ttg/parsec/ttg.h
```diff
@@ -1450,36 +1484,98 @@ namespace ttg_parsec {
...
+        //std::cout << "device_static_op: task " << parsec_task << std::endl;
+
+        /* set up a device task */
+        parsec_gpu_task_t *gpu_task;
+        /* PaRSEC wants to free the gpu_task, because F***K ownerships */
+        gpu_task = static_cast<parsec_gpu_task_t*>(std::calloc(1, sizeof(*gpu_task)));
+        PARSEC_OBJ_CONSTRUCT(gpu_task, parsec_list_item_t);
+        gpu_task->ec = parsec_task;
+        gpu_task->task_type = 0; // user task
+        //gpu_task->load = 1;    // TODO: can we do better?
```
abouteiller, line 1501: the load will be filled up by the default-time-estimate function, which by default will use the compute capacity of the target device to compute in normalized units: 1 load on the fastest device, x load on slower devices where x is the performance ratio of that device w.r.t. the fastest device. In short you should not need to do this commented out thing at all. You can provide better time_estimate functions (see how it's done in ptg) if you want to be more precise w.r.t. the amount of flops individual tasks incur.

ttg/ttg/parsec/device.h
devreal: I still need the first device ID for the conversion between TTG device IDs (matching the CUDA/HIP/L0 ID space) and parsec's ID space. What is the best way for doing that without knowing the first device ID?

ttg/ttg/parsec/ttg.h
devreal: Will be fixed, thanks!

ttg/ttg/parsec/ttg.h
devreal: I removed the comment.

ttg/ttg/parsec/device.h
abouteiller: I don't think we have something nice for that atm. 

BTW you may still want to cache the result because context_query is running a loop.

evaleev (approved):

devreal:
This PR now also includes detection of broken coro support in GCC and fixes for building without coroutine support. I think this is ready for review/shipment.

@abouteiller We'll fix the device ID computation in a separate PR.

abouteiller (approved):

PR #286, opened 2024-06-04
Add static assert for task return and check for derived of ttg::coroutine_handle

devreal:
TTs may return ttg::coroutine_handle (make_tt) or ttg::device::Task (custom TTs)


evaleev (approved):

therault (approved):
Looks good and beautiful to me

PR #293, opened 2024-06-21
[WIP] Add ping-pong test

devreal:
Include both intra- and inter-process ping pong.

We will use this to investigate device pinning by pinning the task instead of the data. This needs an some changes in the parsec backend.

### DIFF
--- a/examples/pingpong.cc
+++ b/examples/pingpong.cc
@@ -1,0 +1,129 @@
+
+#include <ttg.h>
+#include <ttg/serialization.h>
+#include <cassert>
+
+#define NITER 10000
+
+#define MAX_SIZE 1024*1024*1024
+
+struct Message : public ttg::TTValue<Message> {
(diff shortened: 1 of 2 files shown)


examples/pingpong.cc
```diff
@@ -0,0 +1,129 @@
...
+    auto end_ts = std::chrono::steady_clock::now();
+    auto duration = std::chrono::duration<double>(end_ts - start_ts);
+    std::cout << "INTRA " << size <<  " B : " << duration << " s : " << size / duration.count() / 1E6 " MB/s" << std::endl;
+  }
+}
+
+int main(int argc, char **argv) {
+  ttg::initialize(argc, argv);
+  // inter-node benchmark
+  run_inter();
```
abouteiller, line 125: run_intra is never called

abouteiller (approved):

PR #297, opened 2024-09-03
Consolidate make_tt_tpl and make_tt

devreal:
We can map `make_tt_tpl` onto the same mechanism we use for `make_tt` by adding a template template parameter that signals whether arguments are passed as tuple or unpacked.

This is an attempt to clean up the implementation and to open the door for device-support in `ttg::make_tt_tpl`.


ttg/ttg/util/finally.h
```diff
@@ -0,0 +1,57 @@
...
+//
+///////////////////////////////////////////////////////////////////////////////
+
+namespace ttg::detail {
+
+    /* finally, taken from GSL: https://github.com/microsoft/GSL */
+
+    // final_action allows you to ensure something gets run at the end of a scope
+    template <class F>
+    class final_action
```
evaleev, line 26: we might want to consider something more fully featured like https://github.com/justusc/deferral

evaleev: at least let's rename to match what will eventually land in C++: https://en.cppreference.com/w/cpp/experimental/scope_exit/scope_exit

devreal: I took the implementation from https://github.com/tandasat/ScopedResource/ and slightly adopted it.

devreal:
@evaleev I replace `finally` with `exit_scope`. I think this is ready to go.

PR #298, opened 2024-10-16
Upstream improvements needed for MRA on devices

devreal:
Set of changes made while implementing the first MRA kernels with device support. More to come but I wanted to get these out.

Summary:

- Allow specifying the scope (allocate/syncin) on buffers. This is useful for data that is first produced on the device to avoid the initial transfer.
- Allowing `const T` on `ttg::Buffer` to specify that the content of the buffer is immutable.
- Device-side bindings for `broadcastk`
- Deferring errors as long as possible if a type is not copyable. We will try to defer the mutating task until all consumers of the object have completed. Only if we find a second mutating task on the same object we have to throw our hands up.
- Introduce `ttg::device::Stream` representing a device stream object.
- Couple of minor fixes

I tried to break down the changes into individual commits for easier review.

### DIFF
--- a/ttg/ttg/buffer.h
+++ b/ttg/ttg/buffer.h
@@ -1,11 +1,13 @@
 #ifndef TTG_BUFFER_H
 #define TTG_BUFFER_H
 
+#include <memory>
+
 #include "ttg/fwd.h"
 
 namespace ttg {
 
-template<typename T, typename Allocator = std::allocator<T>>
+template<typename T, typename Allocator = std::allocator<std::decay_t<T>>>
 using Buffer = TTG_IMPL_NS::Buffer<T, Allocator>;
 
 } // namespace ttg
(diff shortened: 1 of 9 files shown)


devreal:
Rebased on top of current master. I need a review on this so I can merge and make progress on other PRs.

therault (approved):
LGTM... 

2 questions, the second is not directly related to this PR:
  - this does not change the PaRSEC dependency. Does this work with the master PaRSEC (or at least commit 58f8f3089eca which is selected in TTG master)?
  - we don't have TTG tests that are compiled by default, so we need to go in the output of tests to see if these .h files compile without a problem.... Should we change that at some point?

devreal:
Yes, this does not yet need a new ParSEC version. That will be a separate PR.

The CI should compile all examples and tests and so all headers are compiled multiple times. Is that what you mean?

therault:
OK, good for me.

For the CI, I think we only compile the libraries in the 'build' step, and
then in the 'test' step, we compile the tests to run them. What I mean is
that we need to go inside the 'test' step to find if there is any warning /
compilation error originating from these .h files... But again, that
shouldn't block this PR.

PR #299, opened 2024-10-28
Towards unified device & host tasks

devreal:
We now template the `ttg::device::Task` on the `ExecutionSpace` so that we can determine whether it's a host or device task based on the space. We can then optimize away the select and kernel-wait suspension points. We could remove the send suspension point but we use coroutines for storing the final sends anyway and we don't have access to the task return type in `ttg::device::send()`.

This allows tasks to be written once for both host and device without duplicating much of the code. Host tasks that are not coroutines will continue to be supported.

### DIFF
--- a/examples/potrf/potrf.h
+++ b/examples/potrf/potrf.h
@@ -97,7 +97,7 @@ namespace potrf {
     auto f_dev = [=, iallocator = std::move(iallocator)]
                     (const Key1& key, MatrixTile<T>&& tile_kk,
                      std::tuple<ttg::Out<Key2, MatrixTile<T>>, ttg::Out<Key2, MatrixTile<T>>>& out)
-                  -> ttg::device::Task<ES> {
+                  -> ttg::CoTask<ES> {
       const auto K = key[0];
 
       /* compute successors before submitting the kernel
@@ -225,7 +225,7 @@ namespace potrf {
 #if defined(ENABLE_DEVICE_KERNEL)
     auto f = [=](const Key2& key, const MatrixTile<T>& tile_kk, MatrixTile<T>&& tile_mk,
                  std::tuple<ttg::Out<Key2, MatrixTile<T>>, ttg::Out<Key2, MatrixTile<T>>, ttg::Out<Key3, MatrixTile<T>>,
-                            ttg::Out<Key3, MatrixTile<T>>>& out) -> ttg::device::Task<ES> {
+                            ttg::Out<Key3, MatrixTile<T>>>& out) -> ttg::CoTask<ES> {
       const int M = key[0];
       const int K = key[1];  // the column equals the outer most look K (same as PO)
 
@@ -377,7 +377,7 @@ namespace potrf {
     using T = typename MatrixT::element_type;
 #if defined(ENABLE_DEVICE_KERNEL)
     auto f = [=](const Key2& key, const MatrixTile<T>& tile_mk, MatrixTile<T>&& tile_kk)
-            -> ttg::device::Task<ES> {
+            -> ttg::CoTask<ES> {
       const int K = key[0];
       const int M = key[1];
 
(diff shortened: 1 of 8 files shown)


devreal:
I should also mention that this could allow us to take multiple callables in `ttg::make_tt` and identify their execution space by the return type, laying the ground work for multi-implementation tasks in the future.

devreal:
> perhaps this is the right time to revise the terminology (e.g. ttg::device::Task<> -> ttg::coTask<>?)

I'm not sure about that. We probably don't want to make coroutine tasks the default. So what are they? For me, these tasks are still device tasks but now we have the ability to execute the same task structure on the host. I agree that the naming is a bit funny though.

therault:
`AsyncTask`? Just to make it independent of the technique/implementation used, and try to convey that it can be used for something else than devices?

That being said, `coTask` (or `co_task` to keep a coroutine-like syntax?) conveys the idea that the task itself is a coroutine, which is maybe more helpful for the user.

evaleev:
> I'm not sure about that. We probably don't want to make coroutine tasks the default. So what are they? For me, these tasks are still device tasks but now we have the ability to execute the same task structure on the host. I agree that the naming is a bit funny though.

I agree we don't want to make `coTask` the default. But it's _extremely_ confusing to refer to them as device task. I would prefer a name that does not refer to device to avoid misleading ppl.

evaleev:
> `AsyncTask`? Just to make it independent of the technique/implementation used, and try to convey that it can be used for something else than devices?

To me the word "task" already implies "asynchrony" of execution, by decoupling statement of what to do from when to do it (i.e. after submitting task there is no way to ensure that work has completed unless you await for the result/completion). So `AsyncTask` sounds a bit redundant to me.

> That being said, `coTask` (or `co_task` to keep a coroutine-like syntax?) conveys the idea that the task itself is a coroutine, which is maybe more helpful for the user.

I like `co_task`.

Perhaps what we need is to make cotasks and tasks more symmetric?
```
using Task = void;

make_tt([]() -> Task, ...)
```

devreal:
Proposal from the EPEXA meeting: `ttg::CoTask<ttg::ExecutionSpace::CUDA>` for CUDA or `ttg::CoTask<ttg::ExecutionSpace::Host>` for host execution.

examples/task-benchmarks/chain-ttg-dev.cc
```diff
@@ -3,13 +3,19 @@
 
 #include "chrono.h"
 
-#if defined(TTG_HAVE_CUDA)
+#if defined(CHAIN_CUDA)
```
abouteiller, line 6: What's the difference between CHAIN_CUDA/HIP and ENABLE_CUDA/HIP?

examples/potrf/potrf.h
```diff
@@ -7,22 +7,19 @@
 #include "util.h"
 #include "../devblas_helper.h"
 
-#if (defined(TTG_ENABLE_CUDA) || defined(TTG_ENABLE_HIP))
+#if (defined(TTG_ENABLE_CUDA) || defined(TTG_ENABLE_HIP) || defined(TTG_ENABLE_DEV_HOST))
```
abouteiller, line 10: Naming is inconsistent, `TTG_ENABLE_CPU` ?

tests/unit/device_coro.cc
```diff
@@ -6,17 +6,47 @@
 
 #include "cuda_kernel.h"
 
+#if defined(TTG_HAVE_CUDA)
```
abouteiller, line 9: shouldn't that be TTG_ENABLE_CUDA, or CHAIN_CUDA?

ttg/ttg/device/task.h
```diff
@@ -11,421 +11,440 @@
...
 
-  template <ttg::Runtime Runtime = ttg::ttg_runtime>
-  inline detail::send_t send(std::size_t i) {
-    auto *terminal_ptr = ttg::detail::get_out_terminal<void, void>(i, "ttg::device::send(i, key, value)");
-    return detail::send_t{detail::send_coro(*terminal_ptr)};
-  }
+    template <typename Key, ttg::Runtime Runtime = ttg::ttg_runtime>
+    inline detail::send_t sendk(std::size_t i, const Key& key) {
+      auto *terminal_ptr = ttg::detail::get_out_terminal<Key, void>(i, "ttg::device::send(i, key, value)");
+      return detail::send_t{detail::sendk_coro(key, *terminal_ptr)};
```
abouteiller, line 262: `key` parameter moves from 2nd to first parameter depending on level.

PR #300, opened 2024-10-29
PaRSEC: check for universal peer access

devreal:
Check whether thinks that all devices can access the memory of their peers. If so, we do not need to push modified data back to the host as PaRSEC will use D2D transfers instead of the host copy. We currently stick to the owner-computes model so data will always be modified on the same device and may only be read on a different device.

### DIFF
--- a/ttg/ttg/parsec/ttg.h
+++ b/ttg/ttg/parsec/ttg.h
@@ -1086,7 +1086,7 @@ namespace ttg_parsec {
         parsec_device_gpu_module_t *gpu_device = (parsec_device_gpu_module_t*)device;
         for (int j = 0; (j < parsec_nb_devices) && all_peer_access; ++j) {
           if (PARSEC_DEV_IS_GPU(device->type)) {
-            all_peer_access &= (gpu_device->peer_access_mask & j);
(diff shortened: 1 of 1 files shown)


ttg/ttg/parsec/ttg.h
```diff
@@ -1075,6 +1077,21 @@ namespace ttg_parsec {
...
+
+    bool all_peer_access = true;
+    /* check whether all GPUs can access all peer GPUs */
+    for (int i = 0; (i < parsec_nb_devices) && all_peer_access; ++i) {
+      parsec_device_module_t *device = parsec_mca_device_get(i);
+      if (PARSEC_DEV_IS_GPU(device->type)) {
+        parsec_device_gpu_module_t *gpu_device = (parsec_device_gpu_module_t*)device;
+        for (int j = 0; (j < parsec_nb_devices) && all_peer_access; ++j) {
+          if (PARSEC_DEV_IS_GPU(device->type)) {
+            all_peer_access &= (gpu_device->peer_access_mask & j);
```
abouteiller, line 1089:
```suggestion
            all_peer_access = all_peer_access && (gpu_device->peer_access_mask & j);
```
&= is bitwise

abouteiller (approved):

PR #301, opened 2024-11-08
Replace buffer tracking with serialization inspection

devreal:
We have been tracking the buffers of structures by having the buffer register itself with some metadata. This is clumsy and error-prone (no support for nesting, for example). Now we use a specialized archive to inspect all members of a value and apply some functor to each buffer (or its internal parsec data).

This is also the first step towards a structure derived from parsec_data_copy_t to better manage the life-time of the host copy. We now only allow shared_ptr for non-owning buffers so some adjustments must be made by users.


devreal:
I think this is ready. However, we can only achieve such introspection using the madness archive. With boost, we would have to specialize traits on the inspector archive, which is not possible because the inspector archive is templated on the callable. Even if we were to type-erase the callable into `std::function` we'd still have to pick the value and allocator type for the Buffer that gets passed in. Not possible, because objects can hold different types of buffers at once. 

I don't think this is an issue. Moving from boost to madness serialization is not hard and users have to touch their data structures anyway to integrate buffers. More reliance on madness though, at least for the time being.

devreal:
Nevermind, not ready yet. This requires https://github.com/ICLDisco/parsec/pull/695/files to land in PaRSEC. Reverted back to draft.

devreal:
CI is passing now. Please review :)

devreal:
CI is passing again. @evaleev let's get this in before something else breaks :)

abouteiller (approved):

PR #303, opened 2024-11-12
Add TTG_USE_[PARSEC|MADNESS] to the ttg-parsec/ttg-mad targets

devreal:
No need for applications to specify these flags manually if they already select the `ttg-parsec` or `ttg-mad` targets.


evaleev:
the only scenario that becomes impossible is to have both runtimes present in same executable (hence same target). We don't support or test it right now, so that's fine, but something to keep in mind.

evaleev (approved):

PR #304, opened 2024-11-12
Flexible device inputs

devreal:
Next set of changes needed for MRA. This change set allows buffers to be empty and const and enables tasks to define a flexible set of device inputs using `ttg::device::Input`, which is a vector of buffers internals provided by the backend implementation.


abouteiller (approved):

therault (approved):
We had a live review during the EPEXA meeting, and approved the PR.

PR #307, opened 2024-11-17
PaRSEC now allows DSLs to free the gpu task

devreal:
We can allocate the GPU task inside the task structure and avoid an extra allocation.


devreal:
This doesn't work as thought. PaRSEC releases the task containing the gpu_task structure before the gpu_task is released so we end up with fields overwritten prematurely.

devreal:
I take back what I said earlier. The error was somewhere else and not in this PR. Ready for review.

cmake/modules/ExternalDependenciesVersions.cmake
```diff
@@ -4,7 +4,7 @@
 set(TTG_TRACKED_VG_CMAKE_KIT_TAG d1b34157c349cf0a7c2f149b7704a682d53f6486)  # provides FindOrFetchLinalgPP and "real" FindOrFetchBoost
 set(TTG_TRACKED_CATCH2_VERSION 3.5.0)
 set(TTG_TRACKED_MADNESS_TAG 93a9a5cec2a8fa87fba3afe8056607e6062a9058)
-set(TTG_TRACKED_PARSEC_TAG 58f8f3089ecad2e8ee50e80a9586e05ce8873b1c)
+set(TTG_TRACKED_PARSEC_TAG a9ab33d8287578c68c0349662352f280bc83e2c0)
```
bosilca, line 7: Use the 4.0 please

ttg/ttg/parsec/ttg.h
```diff
@@ -1494,87 +1494,77 @@ namespace ttg_parsec {
...
-
-        parsec_task_class_t& tc = task->dev_ptr->task_class;
-
-        // input flows are set up during register_device_memory as part of the first invocation above
-        for (int i = 0; i < MAX_PARAM_COUNT; ++i) {
-          tc.in[i]  = gpu_task->flow[i];
-          tc.out[i] = gpu_task->flow[i];
+      /* set up the device task */
+      parsec_gpu_task_t *gpu_task = task->dev_ptr->gpu_task;
+      /* TODO: needed? */
```
bosilca, line 1509: You should not need this, you construct the `list_item` and then set the rest of the `gpu_task` fields to default values.

ttg/ttg/parsec/ttg.h
```diff
@@ -1494,87 +1494,77 @@ namespace ttg_parsec {
...
+
+      // first invocation of the coroutine to get the coroutine handle
+      static_op<Space>(parsec_task);
+
+      /* when we come back here, the flows in gpu_task are set (see register_device_memory) */
+
+      parsec_task_class_t& tc = task->dev_ptr->task_class;
+
+      // input flows are set up during register_device_memory as part of the first invocation above
+      for (int i = 0; i < MAX_PARAM_COUNT; ++i) {
```
bosilca, line 1534: Why is the upper bound here always `MAX_PARAM_COUNT` ?

cmake/modules/ExternalDependenciesVersions.cmake
devreal: Too many things missing that we need in PaRSEC so that would last for 1 PR:

- https://github.com/ICLDisco/parsec/pull/710
- https://github.com/ICLDisco/parsec/pull/706
- https://github.com/ICLDisco/parsec/pull/695
- https://github.com/ICLDisco/parsec/pull/687 (or related)
- https://github.com/ICLDisco/parsec/pull/681
- https://github.com/ICLDisco/parsec/pull/671

Maybe 4.1 will work for us.

ttg/ttg/parsec/ttg.h
devreal: Because we don't know how many device inputs the application will give us. We could put a stop there but the impact will be marginal.

devreal:
This PR now depends on https://github.com/ICLDisco/parsec/pull/694 going into PaRSEC and needs some more testing

PR #308, opened 2024-11-19
Pass the Execution Space as TT template parameter

devreal:
Defaults to Host execution so existing code is not affected. Properly set by make_tt.

We cannot query TT::derivedT for flags because at the time TT is instantiated because derivedT is incomplete at that point. For now pass the Space as template parameter. We need to find a different way if we want to have multiple implementations of a task.


devreal:
Please review, this is needed for #301 to pass CI.

evaleev (approved):

PR #311, opened 2024-11-22
Fix check for peer access.

devreal:
Yes, there was a bug. Yes, PaRSEC says devices cannot access themselves. This code should be trivial. Alas...

### DIFF
--- a/ttg/ttg/parsec/ttg.h
+++ b/ttg/ttg/parsec/ttg.h
@@ -1088,12 +1088,15 @@ namespace ttg_parsec {
     bool all_peer_access = true;
     /* check whether all GPUs can access all peer GPUs */
     for (int i = 0; (i < parsec_nb_devices) && all_peer_access; ++i) {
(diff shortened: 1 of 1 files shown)


abouteiller:
I have been considering changing the mask in parsec to have self-access always be true, because that also helps with detecting if the peer-access masks are symmetrical.

devreal:
Yes please. That is the only sane semantic. Ideally, I wouldn't even have to figure it out myself but just ask PaRSEC...

abouteiller:
That being said, why was having self-access causing an actual issue in the end? You should never issue D2D transfers from-to the same device anyways

devreal:
> That being said, why was having self-access causing an actual issue in the end? You should never issue D2D transfers from-to the same device anyways

Because I want to know whether all devices can access each other's memory. If they don't, we have to pushout to the host. Not having the self flag set makes the check more complex and bit me during tests.

PR #312, opened 2024-11-22
Remove OpenMP dependency

devreal:
No code is actually using OpenMP. Some examples still have OpenMP commented out. I'll leave it there for historical reasons.


evaleev (approved):
NB library algebra may inject dependency on openmp

PR #316, opened 2025-02-04
Receive into device

devreal:
We already supported sending from the device. This PR adds support for allocating device buffer space and receiving directly into the device. We do that if we can detect directly whether device memory is supported (OMPI) or if the user tells us that it's safe (`TTG_FORCE_DEVICE_COMM=1`, e.g., with Cray MPICH).

### DIFF
--- a/ttg/ttg/buffer.h
+++ b/ttg/ttg/buffer.h
@@ -71,10 +71,8 @@ namespace madness {
       : m_fn(fn)
       { }
 
-      /// Stores (counts) data into the memory buffer.
 
-      /// The function only appears (due to \c enable_if) if \c T is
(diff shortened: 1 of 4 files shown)


devreal:
With some trial and error I got this to work but I don't think I fully understand PaRSEC's state machine. @bosilca @abouteiller @therault Is it necessary to [set the coherency state](https://github.com/TESSEorg/ttg/pull/316/files#diff-a1666cdfbe57e6602ea3e4f1d000434fcfa72dfd86812b7c876b1911d55bccf9R299) when setting the owning device? I need to set up the data to ensure that PaRSEC picks the device data that I have set...

PR #320, opened 2025-05-20
`umpire-cxx-allocator` is a required prerequisite, provides Umpire allocators without the need to bring in TiledArray

evaleev:
TiledArray is still used as a dependency for building SPMM (because it needs BLAS, BTAS, and Eigen) but madness runtime no longer needs to be initialized for the parsec backend just for the purpose of creating Umpire pinned allocators.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -67,7 +67,7 @@ jobs:
         wget https://developer.download.nvidia.com/compute/cuda/repos/wsl-ubuntu/x86_64/cuda-keyring_1.1-1_all.deb
         sudo dpkg -i cuda-keyring_1.1-1_all.deb
         sudo apt update
-        sudo apt install -y cuda-toolkit
+        sudo apt install -y cuda-toolkit-12
         echo "CUDA_BUILD_OPTS=-DENABLE_CUDA=ON -DTTG_ENABLE_CUDA=ON -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda-12 -DCMAKE_CUDA_COMPILER=/usr/local/cuda-12/bin/nvcc -DCMAKE_CUDA_HOST_COMPILER=${{ matrix.cxx }}" >> $GITHUB_ENV
 
     - name: Create Build Environment


ttg/ttg/env.h
```diff
@@ -0,0 +1,139 @@
...
+
+    inline static std::unique_ptr<Env>& instance_accessor() {
+      static std::unique_ptr<Env> instance_{nullptr};
+      return instance_;
+    }
+  };  // class Env
+
+  namespace umpire {
+
+    struct pinned_allocator_getter {
```
devreal, line 125: Does this really have to be injected into the `umpire` namespace?

ttg/ttg/env.h
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /// @return an Umpire allocator that allocates from a
+    ///         pinned memory pool
+    /// @warning this is not a thread-safe allocator, should be only used when
+    ///          wrapped into umpire_based_allocator_impl
+    auto& pinned_allocator() { return pinned_allocator_; }
+
+    // clang-format off
+  /// @return the max actual amount of memory held by pinned_allocator()
```
devreal, line 96: Indentation is off

CMakeLists.txt
```diff
@@ -184,9 +184,13 @@ if (TTG_HAVE_CUDA OR TTG_HAVE_HIP OR TTG_HAVE_LEVEL_ZERO)
...
-##########################
-#### prerequisite runtimes
-##########################
+############################
+#### mandatory prerequisites
+############################
+add_custom_target(ttg-imported-packages-build)
+# umpire-cxx-allocator
+include(FindOrFetchUmpireCXXAllocator)
+add_dependencies(ttg-imported-packages-build vrg-build-external-projects)
```
devreal, line 193: Should we make this optional and only enable if we're building with CUDA/HIP/L0 support?

devreal (commented):
Thanks @evaleev, just some minor comments.

ttg/ttg/env.h
evaleev, line 96:
```suggestion
      /// @return the max actual amount of memory held by pinned_allocator()
```

cmake/modules/FindOrFetchUmpireCXXAllocator.cmake
```diff
@@ -0,0 +1,39 @@
...
+  endif()
+  if (TTG_ENABLE_HIP)
+    set(UMPIRE_ENABLE_HIP ON CACHE BOOL "Enable HIP support in Umpire")
+  endif()
+
+  include(FetchContent)
+  FetchContent_Declare(
+      umpire-cxx-allocator
+      GIT_REPOSITORY      https://github.com/ValeevGroup/umpire-cxx-allocator.git
+      GIT_TAG             ${TTG_TRACKED_UMPIRE-CXX-ALLOCATOR_TAG}
```
evaleev:
```suggestion
      GIT_TAG             ${TTG_TRACKED_UMPIRE_CXX_ALLOCATOR_TAG}
```

PR #321, opened 2025-05-23
PaRSEC: Defer handling of messages if termination detector is not ready yet

devreal:
We have to make sure that the taskpool is enabled before we allow anyone to send us messages, otherwise PaRSEC's termination detector will bark at us.


devreal:
@evaleev @therault I reverted my first attempt (adding a barrier) and changed the code so that messages received before the TDM is ready are deferred, just like we do if the message handler has not yet been installed. Tests in a loop pass locally, waiting for CI.

evaleev (approved):
I presume there is magic for cleanup of the data `malloc`ed on line 215 ...

evaleev (approved):
Ok thanks for the unique_ptr cleanup

PR #335, opened 2026-02-10
[cmake] autodetect if need additional linker flags to link against libc++

evaleev:
… such as when using Homebrew LLVM


devreal (commented):
It looks like we're pulling in kit-cmake.git for the boost integration as well. Should that be combined into a separate cmake module?

devreal (approved):

